Issue #15, opened 2017-12-31
Header and Footer are not getting printer during conversion of a banner file using bannertopdf.


sahilarora535:
The `Header` and `Footer` keywords in a banner file are not getting printed when converted using `bannertopdf` filter. Here are the contents of the banner file:
```
#PDF-BANNER
Template default.pdf
Header Hello
Footer World
Show job-id
```
The file `default.pdf` can be found [here](https://github.com/OpenPrinting/cups-filters/blob/master/data/default.pdf).

The resulting PDF file do have `job-id`, or other information which is specified in the `Show` keyword, but no header or footer.

Here is the log for the conversion:
```
DEBUG: argv[0]="cupsfilter"
DEBUG: argv[1]="1"
DEBUG: argv[2]="sahil"
DEBUG: argv[3]="default"
DEBUG: argv[4]="1"
DEBUG: argv[5]=""
DEBUG: argv[6]="default"
DEBUG: envp[0]="<CFProcessPath>"
DEBUG: envp[1]="CONTENT_TYPE=application/vnd.cups-pdf-banner"
DEBUG: envp[2]="CUPS_DATADIR=/usr/share/cups"
DEBUG: envp[3]="CUPS_FONTPATH=/usr/share/cups/fonts"
DEBUG: envp[4]="CUPS_SERVERBIN=/usr/lib/cups"
DEBUG: envp[5]="CUPS_SERVERROOT=/etc/cups"
DEBUG: envp[6]="LANG=en_US.UTF8"
DEBUG: envp[7]="PATH=/usr/lib/cups/filter:/usr/bin:/usr/sbin:/bin:/usr/bin"
DEBUG: envp[8]="PPD=/usr/share/cups/model/laserjet.ppd"
DEBUG: envp[9]="PRINTER_INFO=cupsfilter"
DEBUG: envp[10]="PRINTER_LOCATION=Unknown"
DEBUG: envp[11]="PRINTER=cupsfilter"
DEBUG: envp[12]="RIP_MAX_CACHE=128m"
DEBUG: envp[13]="USER=sahil"
DEBUG: envp[14]="CHARSET=utf-8"
DEBUG: envp[15]="FINAL_CONTENT_TYPE=application/pdf"
INFO: bannertopdf (PID 22483) started.
DEBUG: Could not open PPD file '/usr/share/cups/model/laserjet.ppd'
PDF template file doesn't have form. It's okay.
[...]

Issue #14, opened 2017-12-30
PDF forms converted using bannertopdf show gibberish text when the form is edited.


sahilarora535:
I created a PDF form by converting a `PDF-BANNER` file. The contents of the banner file are as follows:
```
#PDF-BANNER
Template form_english.pdf
```
The file `form_english.pdf` can be found [here](https://github.com/OpenPrinting/cups-filters/blob/master/data/form_english.pdf).

The banner file is converted to PDF using the following command:
```
cupsfilter banner_file > OUT.pdf
```

In the original file, if a field in a form is edited, it gets correctly edited and is reflected in the file. However, in the converted PDF file `OUT.pdf`, the form fields in the file get filled with gibberish as soon as the field is clicked on, or if tried to be edited. See here sample image. The field 'Task name' was edited and this is the final result:

![image](https://user-images.githubusercontent.com/12633577/34456564-900b3ad4-edbf-11e7-9ee3-b829ad8e26d5.png)

Here is the log for the conversion:

```
DEBUG: argv[0]="cupsfilter"
DEBUG: argv[1]="1"
DEBUG: argv[2]="sahil"
DEBUG: argv[3]="default"
DEBUG: argv[4]="1"
DEBUG: argv[5]=""
DEBUG: argv[6]="default"
DEBUG: envp[0]="<CFProcessPath>"
DEBUG: envp[1]="CONTENT_TYPE=application/vnd.cups-pdf-banner"
DEBUG: envp[2]="CUPS_DATADIR=/usr/share/cups"
DEBUG: envp[3]="CUPS_FONTPATH=/usr/share/cups/fonts"
DEBUG: envp[4]="CUPS_SERVERBIN=/usr/lib/cups"
DEBUG: envp[5]="CUPS_SERVERROOT=/etc/cups"
DEBUG: envp[6]="LANG=en_US.UTF8"
DEBUG: envp[7]="PATH=/usr/lib/cups/filter:/usr/bin:/usr/sbin:/bin:/usr/bin"
DEBUG: envp[8]="PPD=/usr/share/cups/model/laserjet.ppd"
DEBUG: envp[9]="PRINTER_INFO=cupsfilter"
DEBUG: envp[10]="PRINTER_LOCATION=Unknown"
DEBUG: envp[11]="PRINTER=cupsfilter"
DEBUG: envp[12]="RIP_MAX_CACHE=128m"
DEBUG: envp[13]="USER=sahil"
[...]

PR #25, opened 2018-02-23
Rewrite bannertopdf to use QPDF instead of Poppler

### LINKED ISSUE #8: bannertopdf should not depend on Poppler
bannertopdf generates PDF files for banner pages and test pages. It does not render existing PDF files, so it does not need a PDF interpreter, so Poppler is not really needed for this task. QPDF would be good enough here. QPDF is a helper to generate and modify PDFs, especially to create a correct XREF table.
Also we want to give the user the liberty to use cups-filters with one single PDF interpreter, one of Ghostscript, Poppler, and MuPDF, so that on devices with low resources (mobile, IoT, ...) one does not need the redundancy of more than one PDF interpreter.
Currently one looses banner and test page functionality if one does not have Poppler installed, so actually only Poppler works as a single PDF interpreter for cups-filters.
It is also possible that bannertopdf uses non-public (internal) APIs of Poppler which can make it easily break with updates of Poppler, but I am not sure whether this is really the case.
Therefore we need to convert bannertopdf to use QPDF instead of Poppler.
Hint: All Poppler use is in filter/pdf.cxx.

sahilarora535:
image

**Introduction**
This pull request caters to the project **["Improving bannertopdf filter to use QPDF instead of Poppler APIs"](https://summerofcode.withgoogle.com/projects/#5640715745361920)** being done as a part of the **[Google Summer of Code Program](https://summerofcode.withgoogle.com/)**, 2018. 

Remove all Poppler code from `bannertopdf` and re-write it using QPDF.

Fixes #8.

Depends on https://github.com/qpdf/qpdf/issues/178 for completion.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -69,15 +69,20 @@ License: Expat
 
 Files: filter/banner.c
   filter/banner.h
-  filter/pdf.h
   banners/*
   data/*
 Copyright: 2012, Canonical Ltd
 License: GPL-3
 
+Files: filter/pdf.h
+Copyright: 2018, Sahil Arora <sahilarora.535@gmail.com>
+  2012, Canonical Ltd
+License: GPL-3
+
 Files: filter/bannertopdf.c
   filter/pdf.cxx
-Copyright: 2013, ALT Linux, Andrew V. Stepanov <stanv@altlinux.com>
+Copyright: 2018, Sahil Arora <sahilarora.535@gmail.com>
+  2013, ALT Linux, Andrew V. Stepanov <stanv@altlinux.com>
   2012, Canonical Ltd
 License: GPL-3
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -119,9 +119,7 @@ bannerfiles = \
 	banners/topsecret \
 	banners/unclassified
 
-if ENABLE_POPPLER
 pkgbanner_DATA += $(bannerfiles)
-endif
 
 EXTRA_DIST += $(bannerfiles)
 
@@ -612,6 +610,7 @@ pkgfilter_PROGRAMS += \
 	rastertopclx \
 	texttopdf \
 	rastertopdf \
+	bannertopdf \
 	rastertops
 if ENABLE_URFTOPDF
 pkgfilter_PROGRAMS += \
@@ -619,7 +618,6 @@ pkgfilter_PROGRAMS += \
 endif
 if ENABLE_POPPLER
 pkgfilter_PROGRAMS += \
-	bannertopdf \
 	pdftoraster
 if ENABLE_IJS
 pkgfilter_PROGRAMS += \
@@ -673,7 +671,7 @@ bannertopdf_CFLAGS = \
 	$(CUPS_CFLAGS) \
 	$(LIBJPEG_CFLAGS) \
 	$(LIBPNG_CFLAGS) \
-	$(POPPLER_CFLAGS) \
+	$(LIBQPDF_CFLAGS) \
 	$(TIFF_CFLAGS) \
 	$(FONTCONFIG_CFLAGS) \
 	-I$(srcdir)/fontembed/
@@ -684,7 +682,7 @@ bannertopdf_LDADD = \
 	$(CUPS_LIBS) \
 	$(LIBJPEG_LIBS) \
 	$(LIBPNG_LIBS) \
-	$(POPPLER_LIBS) \
+	$(LIBQPDF_LIBS) \
 	$(TIFF_LIBS) \
 	$(FONTCONFIG_LIBS) \
 	libfontembed.la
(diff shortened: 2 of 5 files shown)


tillkamppeter:
@sahilarora535, QPDF 8.0.2 is released. Does it already contain everything to finish bannertopdf?

sahilarora535:
@tillkamppeter QPDF 8.1.0 has been released, and it contains most of the features required by us. For some of the features like changing font in a form field, I need to see whether we can do it with the current set of features available. I will update you as soon as I am done with the analysis. Until then, I am completing this branch and finalizing it for merging with master branch. I hope that makes sense.

tillkamppeter:
OK, Let us go this way.

sahilarora535:
@tillkamppeter I have pushed more changes to this branch. The status of the work as of now is:
1. The basic functionality of `bannertopdf` is complete. The newly-written filter is able to process forms, fill them with the correct values and produce correct PDFs. It detects whether there is a form or not, and does its work accordingly. It would be very helpful if you could also test it to see if things are according to what we need. The code can be tested using the scripts here: https://github.com/sahilarora535/bannertopdf-test
All you need to do is compile the source on this branch and run the script [`convert`](https://github.com/sahilarora535/bannertopdf-test/blob/master/convert). Basically, I have successfully been able to replace all functionality of Poppler which was being used in `bannertopdf`with QPDF, tested the functionality, and I think this can enable us to get this new filter in Ubuntu 18.10. 
2. The issue raised in #14 has been fixed automatically after replacing Poppler with QPDF.
3. The issue raised in #15 has not been fixed, because essentially in the previous version of `bannertopdf`, header and footer were never even written to the PDF during conversion. I'll have to look into how to write them to the banner file.
4. In the previous version of `bannertopdf` which used Poppler, the `Font-Size` and `Font-Face` were being taken as input, but I could not find `Font-Size` being updated to the PDF, nor was the `Font-Face` being added correctly. I tried changing these attributes but could not see any visible change. The part which deals with changing the font-face looks a bit complicated and it might take some time to understand. There is a lot of font conversions, such as UTF-8 to UTF-16, latin characters, etc., most of which I have very little or almost no idea about. I do not currently know how to do that using QPDF, but am looking into it.
5. I have removed all the conflicts of this branch with master branch so that it can be merged.

Once you test this code and it is upto the mark, I can then refactor it a bit and squash the commits to give it the final touch and then merge it with master. Then we can work on other enhancements as I have mentioned above, since points 3 and 4 above are no breaking changes, just enhancements to `bannertopdf`. Or should I try to fix everything and then push it to master? WDYT on this?

tillkamppeter:
I have tested your QPDF-based `bannertopdf` using your `convert` test script and it works so far. Especially issue #14 is fixed.

What does not work is the logging. All widget names in the log output come out as gibberish, for example:
```
Fill widget name ��tV�U with value N/A.
```
This needs to be fixed, not only to make the logging useful but also to avoid that `bannertopdf` accesses arbitrary memory which it is not supposed to access, as this can cause security problems.
Also log messages of CUPS filters must be CUPS-conforming, usually starting with "DEBUG: ", "DEBUG2:", ...

sahilarora535:
@tillkamppeter I have fixed the issue you mentioned. Thanks a lot for noting them. You can test the changes again. Please let me know if you face any other issues.

tillkamppeter:
Thank you very much. It is working correctly now.

tillkamppeter:
If you fix everything and push it to master would be great, if you have the time and the fixes are not too complicated. Otherwise if it makes the originally intended features working and has no regressions against the Poppler version you could prepare it for merging upstream.

jberkenbilt:
I haven't digested your full message, but I wanted to mention that qpdf 8.1
added some methods for dealing with Unicode strings. They were added
specifically for forms, but they work in other places too. Search for
"UTF-16" in QPDFObjectHandle.hh. The method
QPDFObjectHandle::newUnicodeString takes a UTF-8 encoded std::string as
input and produces a QPDF Object that outputs as a UTF-16-encoded PDF
string. None of the examples use this, but you can look
at qpdf/test_pdf_unicode.cc to see an example of its use. Mostly you don't
have to worry about it, but if you are trying to set form field values to
Unicode strings that are not representable with PDFDocEncoding, this is
what you need.

sahilarora535:
> I haven't digested your full message, but I wanted to mention that qpdf 8.1 added some methods for dealing with Unicode strings. They were added specifically for forms, but they work in other places too. Search for "UTF-16" in QPDFObjectHandle.hh. The method QPDFObjectHandle::newUnicodeString takes a UTF-8 encoded std::string as input and produces a QPDF Object that outputs as a UTF-16-encoded PDF string. None of the examples use this, but you can look at qpdf/test_pdf_unicode.cc to see an example of its use. Mostly you don't have to worry about it, but if you are trying to set form field values to Unicode strings that are not representable with PDFDocEncoding, this is what you need.

@jberkenbilt Thank you very much for your comment on unicode strings and the example file really helped. The previous version of the filter did have a separate routine for converting UTF8 to UTF16.

Also @tillkamppeter the pull request is complete and ready to merge from my side. It would be very helpful if you could review it once and merge it if all is good.

PR #49, opened 2018-06-12
Revert "driverless, foomatic-rip: Create relative symbolic links."

somasis:
This reverts commit 0297a70022e366c56e3c8d4b1036df80874d8bb4.

The `-r`/`--relative` flag in `ln` is a GNU extension, and this commit adds a
somewhat unnecessary dependency on GNU coreutils. Reverting this allows for
building with other coreutils implementations, like Busybox.


tillkamppeter:
For distribution packaging and for sandbox setups (like Snaps) relative symlinks are important here. Do you have an idea how to easily generate them without `-r` option?

somasis:
Unfortunately, without knowing the directories or walking the 
directories in shell to construct a path, there isn't a way to replace 
it. At least not a (clean and reasonable) POSIX method...

tillkamppeter:
Fixed by pull request #154

PR #53, opened 2018-08-15
Fixing covscan issues

zdohnal:
Hi Till,
I ran static analysis by coverity scan for cups-filters-1.20.0, which we have in Fedora 28, and fixed several issues which were reported and they are even in current version. I'll send you initial results of scan by email (in case there are some issues, which can be exploitable).
Does it look good to you? Would you mind adding it to the project?

### DIFF
--- a/backend/beh.c
+++ b/backend/beh.c
@@ -223,6 +223,8 @@ call_backend(char *uri,                 /* I - URI of final destination */
   */
 
   strncpy(scheme, uri, sizeof(scheme));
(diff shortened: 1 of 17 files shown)


tillkamppeter:
@zdohnal, thank you very much for doing the tests and providing the patches. I have merged your patches now.

zdohnal:
Np, all issues in report which sent you were counted as important, so there are other issues, but coverity doesn't count them as important.

PR #71, opened 2018-11-14
Keep actually current time for timeouts

zdohnal:
Hi Till,

I have a report in Fedora https://bugzilla.redhat.com/show_bug.cgi?id=1648697 , where cups-browsed time outs on connection to local CUPS before finishing creation of queue and causes high CPU load on cupsd process.
 I am aware of httpLocalTimeout directive, which can set timeout for the connection and it fixes the issue, but I'm concerned about how fast retry of queue creation happens - TIMEOUT_RETRY is set 10s, but the retry happens in 3s or 4s after reaching timeout, which causes high load in CUPS.

It is caused by incorrect current time value in update_printer_queues(), which is defined at the start of function, and never refreshed after that, so it might happen cups-browsed floods cupsd with requests, because intervals between retries are short, because daemon spent some amount of time in the function and setting next timeout for retry does not reflect it.

I ran 'make check' with the patch and it ended without errors.

If the user confirms it fixed the issue, would you mind adding the patch to the project?

### DIFF
--- a/utils/cups-browsed.c
+++ b/utils/cups-browsed.c
@@ -4407,7 +4407,7 @@ gboolean update_cups_queues(gpointer unused) {
   int num_jobs;
   cups_job_t *jobs;
   ipp_t *request;
-  time_t current_time = time(NULL);
+  time_t current_time;
   int i, new_cupsfilter_line_inserted, ap_remote_queue_id_line_inserted,
     cont_line_read, want_raw;
   char *disabled_str, *ptr, *prefix;
@@ -4461,6 +4461,9 @@ gboolean update_cups_queues(gpointer unused) {
   for (p = (remote_printer_t *)cupsArrayFirst(remote_printers);
        p; p = (remote_printer_t *)cupsArrayNext(remote_printers)) {
 
+    /* We need current time as precise as possible because of timeouts for possible retries */
+    current_time = time(NULL);
+
     /* terminating means we have received a signal and should shut down.
        in_shutdown means we have exited the main loop.
        update_cups_queues() is called after having exited the main loop
@@ -4503,6 +4506,7 @@ gboolean update_cups_queues(gpointer unused) {
 	if ((http = http_connect_local ()) == NULL) {
 	  debug_printf("Unable to connect to CUPS!\n");
 	  if (in_shutdown == 0)
+            current_time = time(NULL);
 	    p->timeout = current_time + TIMEOUT_RETRY;
 	  break;
 	}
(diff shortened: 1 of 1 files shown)


tillkamppeter:
Thank you for the patch. Please tell me as soon as the user has confirmed and I will merge the fix.

zdohnal:
Hi Till,
timeouts work better after the fix (cupsd doesn't eat all CPU), I currently work on setting some retries for every queue, so the ones, which get timeouts, will be skipped after several retries (they will be skipped forever until cups-browsed restart or OS restart). Do you think it is acceptable behavior for cups-browsed and could be added to the project?
If you are okay with it , then I'll keep you updated when I'm done (and I'll open separate pull request for it ).

zdohnal:
@tillkamppeter Till, actually timeouts during updating queues will need one adjustment in recheck_timer():
---------------------------------------------
 @@ -5109,7 +5166,7 @@ recheck_timer (void)
      else if (now > p->timeout) {
        timeout = 0;
        break;
 -    } else if (timeout == (time_t) -1 || p->timeout - now < timeout)
 +    } else if (timeout == (time_t) -1 || p->timeout - now > timeout)
        timeout = p->timeout - now;
 
    if (queues_timer_id)
------------------------------------------------

because if p->timeout for queue 1 is already due (now >= queue1->timeout) and queue 2 should be retried after timeout in 10s then the function will choose timeout for queue 1 during evaluation in recheck_timer() and runs update_cups_queues for all remote printers.
IMO cups-browsed should choose the biggest p->timeout available (as it is suggested in the patch) or call update_cups_queues() for each printer separately (which would need rewriting a lot of code) to honor print queue timeouts.

tillkamppeter:
Your patch is wrong. recheck_timer() determines in how many seconds it is the next time when at least one queue needs attention. After this amount of seconds elapsing update_cups_queues() is called. This does not mean that all queues get treated this time. update_cups_queues() by itself checks the queue's timeouts and only treats the queues whose timeouts are already elapsed. So there is no early or additional treatment of any queue.
With your change it finds (from the queues which need attention currently) the queue whose turn is the farthest away, and all other queues have to wait until then. Only exception is if one queue's timeout is already elapsed. Then update_cups_queues() is called immediately.

zdohnal:
Yes, you're right - I missed if (p->timeout > current_time) in every case statement. Sorry for the noise - I was confused by several messages about updating queues in 0s, which were caused by setting queue timeout to X and then creation of other queue timeouted - and timeout for first queue elapsed during it, so recheck_timer() set next update_cups_queues() to 0s...

PR #72, opened 2018-11-22
Call update_cups_queues() after the biggest timeout available

zdohnal:
Patch for the issue mentioned in #71 . Till, do you think it is okay?

### DIFF
--- a/utils/cups-browsed.c
+++ b/utils/cups-browsed.c
@@ -5493,7 +5493,7 @@ recheck_timer (void)
     else if (now > p->timeout) {
       timeout = 0;
       break;
-    } else if (timeout == (time_t) -1 || p->timeout - now < timeout)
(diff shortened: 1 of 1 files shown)


tillkamppeter:
Your patch is wrong. recheck_timer() determines in how many seconds it is the next time when at least one queue needs attention. After this amount of seconds elapsing update_cups_queues() is called. This does not mean that all queues get treated this time. update_cups_queues() by itself checks the queue's timeouts and only treats the queues whose timeouts are already elapsed. So there is no early or additional treatment of any queue.
With your change it finds (from the queues which need attention currently) the queue whose turn is the farthest away, and all other queues have to wait until then. Only exception is if one queue's timeout is already elapsed. Then update_cups_queues() is called immediately.

PR #83, opened 2018-12-29
Poppler 0 72

StefanBruens:
This series fixes building of pdftoopvp with poppler up to 0.72.


tillkamppeter:
Thank you very much for the patches. I have merged them now.
Are you actually using pdftoopvp? If so, for which printer and which driver?

StefanBruens:
> Are you actually using pdftoopvp? If so, for which printer and which driver?

No, but I got a message from an apparent user yesterday.

tillkamppeter:
Could you ask that user?

tillkamppeter:
@mpsuzuki, are these drivers supplied by NEC and Canon? Or by a third party? For which printer models are they? Are these Japan-(China, Korea, Eastern Asia)-only? Or are these printers also available in Europe or the US? If so, are they also using OPVP in their US/European driver versions?
Where can I download these drivers?

mpsuzuki:
Like this one:
https://jpn.nec.com/printer/laser/support/os/linux/download/npdl-driver.html
you can find the statement "本ドライバはOpenPrinting Workgroupが標準化を進めるVector Printer Driver APIに基くプリンタドライバです。".

tillkamppeter:
These drivers look very old, with the newest supported Ubuntu version being 12.04 (April 2012). Are they really still maintained? Do current NEC printer models still use these drivers? Or do they use more modern PDLs, like PDF?

StefanBruens:
For the NEC MultiWriter, there is support using Ghostscripts npdl driver (PDF -> foomatic-rip/GS -> npdl). The other option would be a rastertonpdl driver, see http://git.ghostscript.com/?p=ghostpdl.git;a=blob;f=contrib/japanese/gdevnpdl.c

Current NEC models seem to directly support IPP, while some older ones support it optionally, see https://translate.google.de/translate?hl=&sl=ja&tl=en&u=https%3A%2F%2Fjpn.nec.com%2Fprinter%2Flaser%2Fsupport%2Fnetwork%2Fnw%2Findex.html

mpsuzuki:
Even if the drivers look old, the printers are still being sold, and the last drivers for Linux are them, and they might be the most possible way to use the following models. For example, NEC MultiWriter 8300 (since 2017) does not provide official Linux support anymore. But their native PDL is still NPDL which the earlier models were using.

How about the case of CANON?
https://cweb.canon.jp/drv-upd/lasershot/linux/captlinux.html
According to the document, to install their drivers, the ghostscript supporting OPVP device is required.

StefanBruens:
> Even if the drivers look old, the printers are still being sold, and the last drivers for Linux are them, and they might be the most possible way to use the following models. For example, NEC MultiWriter 8300 (since 2017) does not provide official Linux support anymore. But their native PDL is still NPDL which the earlier models were using.

Should be possible to use it via ghostscript npdl then ...

> How about the case of CANON?
> https://cweb.canon.jp/drv-upd/lasershot/linux/captlinux.html
> According to the document, to install their drivers, the ghostscript supporting OPVP device is required.

Some of Canon's printers support the LIPS PDL, which is supported via ghostscript. Current laser printers sold in Europe (probably the same ones with different marketing names) typically support PS and/or PCL5/PXL.

mpsuzuki:
LIPS is for hi- or mid-range printers, lo-end CANON printers do not support them.
https://cweb.canon.jp/satera/lbp/comparison/pdf/sateralbp-spec.pdf
you can find some models do not have LIPS but CAPT or CARPS2.

StefanBruens:
> LIPS is for hi- or mid-range printers, lo-end CANON printers do not support them.
> https://cweb.canon.jp/satera/lbp/comparison/pdf/sateralbp-spec.pdf
> you can find some models do not have LIPS but CAPT or CARPS2.

LBP 251/252, which are shown as "CARPS2", both support PCL/PXL. There are many other printers from the LBP line which support PS/PCL/PXL.

mpsuzuki:
Oh, I was not aware of that LBP 251/252 supports PCL. Thank you very much (I could not find such infos in their catalogues).
However, the drivers provided by CANON do not seem to emit PCL. And, it is told that CANON's CAPT or CARPS2 printers should be connected via CANON's proprietary connection daemon "ccpd".
The people, who use LBP 251/252 with PCL, are connecting via ccpd? Or, they can do that without relying any CANON's drivers?

Issue #22, opened 2018-02-15
Can't add Canon MG4120 anymore (cups-filters 1.20.0)


P3p1t0:
I use Print Setting ([system-config-printer](https://github.com/zdohnal/system-config-printer)) in order to add printers.

When I try to add my printer by automatic detection, it just fails with "server-error-internal-error". Normally, it should add my printer and use driverless for printing. This GUI tool always chooses driverless by default when you use automatic detection; it doesn't let you choose between driverless and something else.

It worked perfectly with previous of cups-filters, including 1.18.0 and 1.19.0. I got this issue when I updated to 1.20.0. I could add my printer and use driverless for it with previous versions of cups-filters.

A workaround I found in order to add my printer with system-config-printer is to manually add the printer, then manually select Gutenprint. Trying to add the printer manually with driverless instead of Gutenprint results in the same error I get when I try to add it by automatic detection.

Trying to add the printer with the Web interface (localhost:631) using driverless ends up with an error too : "unable to add error : unable to copy PPD file." However, adding the printer with the Web interface using Gutenprint works fine.

Here's a typical log output from /var/log/cups/error_log I get when I try to add my printer with driverless in general.

```
E [04/Feb/2018:20:58:30 -0500] [cups-deviced] PID 5959 (gutenprint52+usb) stopped with status 1!
E [04/Feb/2018:20:58:44 -0500] [CGI] Unable to create PPD file: Printer does not support required IPP attributes or document formats.
E [04/Feb/2018:20:58:47 -0500] [CGI] Unable to create PPD file: Printer does not support required IPP attributes or document formats.
E [04/Feb/2018:20:58:47 -0500] copy_model: empty PPD file
E [04/Feb/2018:20:58:47 -0500] [Client 118] Returning IPP server-error-internal-error for CUPS-Add-Modify-Printer (ipp://localhost/printers/Canon-MG4100-series) from localhost
E [04/Feb/2018:20:58:47 -0500] [CGI] Unable to create PPD file: Printer does not support required IPP attributes or document formats.
E [04/Feb/2018:20:58:47 -0500] copy_model: empty PPD file
E [04/Feb/2018:20:58:47 -0500] [Client 122] Returning IPP server-error-internal-error for CUPS-Add-Modify-Printer (ipp://localhost/printers/Canon-MG4100-series) from localhost
```
Hope you can do something about it.

**Reported about the issue on Manjaro forum**
[...]

PR #86, opened 2019-01-06
cups-browsed, driverless: request urf-supported attribute

mark246:
Fixes bug #22 - at least for me (I have a Canon Pixma MG4250 printer).

I have not tested other Canon models, so can't be certain that this will fix it for other Pixma variants referred to in the bug report (and also can't be sure that requesting this attribute won't be a breaking change for other printers).

### DIFF
--- a/utils/cups-browsed.c
+++ b/utils/cups-browsed.c
@@ -3849,7 +3849,8 @@ get_printer_attributes(const char* uri) {
     "job-template",
     "printer-defaults",
     "printer-description",
-    "media-col-database"
+    "media-col-database",
+    "urf-supported"
   };
   /* Request printer properties via IPP to generate a PPD file for the
      printer (mainly driverless-capable printers)
--- a/utils/driverless.c
+++ b/utils/driverless.c
@@ -482,7 +482,8 @@ generate_ppd (const char *uri)
     "job-template",
     "printer-defaults",
     "printer-description",
-    "media-col-database"
+    "media-col-database",
+    "urf-supported"
   };
 
   /* Request printer properties via IPP to generate a PPD file for the


mark246:
[cups-filters-issue-22-notes.txt](https://github.com/OpenPrinting/cups-filters/files/2730975/cups-filters-issue-22-notes.txt)

mark246:
@tillkamppeter hello and thanks for taking the time to look at my pull request  (my first public one on github). If there are any alternative approaches to fixing this you'd like me to try or you need more logs etc then I'll be happy to help.

tillkamppeter:
cups-filters 1.19.0 did mot add a "requested-attributes" attribute to the get-printer-attributes request, which, unfortunately, did not lead to all attributes being returned. Especially the "media-col-database" was not returned and so not all media sizes and types got reported. Therefore I introduced this attribute in commit 4b92fffecde on Jan 26, 2018, copying from CUPS' `lpadmin` utility (see below).

I have checked CUPS. the `lpadmin` utility (with "`-m everywhere`, file `systemv/lpadmin.c`) does exactly the same as cups-browsed and driverless, so has the same bug, but the scheduler (CUPS daemon) does different when auto-generating print queues for IPP printers.
Instead of a line
```
ippAddStrings(request, IPP_TAG_OPERATION, IPP_TAG_KEYWORD,
                 "requested-attributes", sizeof(pattrs) / sizeof(pattrs[0]),
                 NULL, pattrs);
```
with appropriately set `pattrs` array (the array where you have added "`urf-supported`" in your patch) it uses
```
ippAddString(request, IPP_TAG_OPERATION, IPP_TAG_KEYWORD,
                 "requested-attributes", NULL, "all");
```
and no need of a `pattrs` array (function `create_local_bg_thread()` in `scheduler/ipp.c`).
Could you edit both `utils/cups-browsed.c` and `utils/driverless.c` to use this method and see whether it works with your printer (reporting "urf-supported") AND reports the "media-col-database"?

By the way, the CUPS web interface ( `get_printer_ppd()` function in `cgi-bin/admin.c`) does not add "requested-attributes" and so would support your printer but not use "media-col-database".

In any case we need to report a bug to CUPS, as it is completely inconsistent, but at first we need to find the correct solution here in cups-filters.

tillkamppeter:
Please also test to still use the `pattrs` array, but instead of the original entries use only the two entries "all" and "media-col-database". Does this also work with your printer and provide the "media-col-database" entry in the IPP response?
This is the method which is used by `ipptool` with the template file `/usr/share/cups/ipptool/get-printer-attributes-2.0.test`.

mark246:
@tillkamppeter I've tried the two alternative solutions you suggested. In each case I captured the IPP request/response like this in each case:

    sudo tshark -i enp2s0  -f "host 192.168.1.118" -Y "ipp" -T json
    make driverless && ./.libs/driverless ipp://004417000000.local:631/ipp/printer 

But here's the summary of the results.

**Attempt 1** (Code same as in my current pull request)

- PPD is generated "successfully" but contains no MediaType information.
- "urf-supported present in response
- "media-col-database" is huge, and nonsensical (contains the same data block repeated over-and-over).

**Attempt 2** (ippAddString, no pattrs, "all")

- PPD is generated, and DOES contain MediaType information
- "printer-attributes-tag" section is present, including "urf-supported" field
- Response is a reasonable size (4K), but "media-col-database" is NOT present

**Attempt 3** (pattrs = {"all", "media-col-database"})

- PPD generation fails, with "Printer does not support required IPP attributes or document formats"
- The response is huge and contains a corrupt media-col-database similar to attempt 1
- "printer-attributes-tag" section is missing, so there's no "urf-supported" field
- Power-cycling the printer and repeating does not "fix" the corrupt media-col-database.

tillkamppeter:
As in CUPS (see [CUPS issue #5484](https://github.com/apple/cups/issues/5484)) I use the attributes "`all`" and "`media-col-database`" now. This lets the "`urf-supported`" and also all page size and margin info get corectly reported.
Fixed in cups-filters with commit 5259bb798.

PR #88, opened 2019-01-13
Use qpdf to flatten forms

smilingthax:
CAVEATS: Will make cups-filter to depend on qpdf 8.4.0, which is not even available yet!
I'm also not sure about the issues to reference...

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -9,6 +9,7 @@ CHANGES IN V1.21.7
 	  Poppler 0.18 (Pull request #83).
 	- cups-browsed: Fixed crash in applying the BrowseFilter
 	  cups-browsed.conf directives (Debian bug #916765).
+	- pdftopdf: Use QPDF for form flattening (Issues #23, #36, #2)
 
 CHANGES IN V1.21.6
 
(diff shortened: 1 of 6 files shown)


tillkamppeter:
Thank you very much for your patches, but my suggestion is to do it differently:
The flattening workaround based on pdftocairo and Ghostscript should not be removed but stay inside as an alternative way of flattening. The QPDF solution will be used by default then and the external utilities users can activate as alternative should they run into problems with the QPDF-based flattening.
Similar to the possibility to choose the PDF interpreter in the pdftops filter using the "pdftops-renderer" (see README file of cups-filters) the form flattening method should be also selectable by an option. My suggestion is to call the option "pdftopdf-form-flattening" with the choices "qpdf", "pdftocairo", and "ghostscript".
In addition, there should be conditional compiling to allow building cups-filters also with QPDF < 8.4.0 but then without QPDF-based form-flattening.
If the "pdftops-renderer" is not supplied and pdftopdf got built with QPDF-based form-flattening, this is used by default, if it was built without QPDF-based form-flattening, the two alternatives are tried, first pdftocairo and then Ghostscript, as before.
Can you implement it this way? Thanks in advance.

tillkamppeter:
The bug with the closing of the infile I have fixed now. Thank you very much for making me aware of this.
I have checked and it is actually possible to support both QPDF-based and external flattening with selection of the method by an option. I will implement this soon.

PR #91, opened 2019-02-18
Do not do IPP request for printer-is-shared option for remote cups qu…

zdohnal:
…eues with CUPS 2.2 and newer

Hi!

I found out there are lots of error messages in journal from CUPS in Fedora 28 (cups-filters-1.20.0, cups-2.2.6), when I run cups-browsed with BrowsePoll directive set on remote CUPS server (RHEL 6, cups-1.4.2) like:

[Client 13] Returning IPP client-error-bad-request for CUPS-Add-Modify-Printer (ipp://localhost:631/printers/print_queue_name) from localhost

It is caused by cups-browsed trying to set printer-is-shared option to false for remote CUPS queue, which is forbidden by CUPS as you wrote in comments in the code.
It is harmless error, but it can mystify an user and log is IMHO nicer without such errors.
I introduced a new macro in cups-browsed.c, HAVE_CUPS_2_2, to indicate if we have cups-2.2 or newer in the fix. Then, in update_cups_queues(), when function is issuing a request for printer-is-shared option, do not request the option when the print queue is remote CUPS queue and we have CUPS-2.2 or newer.

Does it look ok? Would it be possible to add it to the project? I tried it on Fedora 28 machine, remote cups queue was found, with no errors in logs, and it prints fine.

Thank you in advance and have a nice day!

Zdenek

### DIFF
--- a/utils/cups-browsed.c
+++ b/utils/cups-browsed.c
@@ -451,6 +451,19 @@ static remote_printer_t
 #endif
 
 /*
+ * Option 'printer-is-shared' cannot be set on remote CUPS
+ * queue and requests for setting it ends with error since
+ * 2.1.1. Define HAVE_CUPS_2_2 to do not send IPP request
+ * for setting 'printer-is-shared' option on remote CUPS queues
+ * for newer versions of CUPS.
+ */
+#if (CUPS_VERSION_MAJOR > 2) || (CUPS_VERSION_MINOR > 1)
+#define HAVE_CUPS_2_2 1
+#else
+#define HAVE_CUPS_2_2 0
+#endif
+
+/*
  * CUPS 1.6 makes various structures private and
  * introduces these ippGet and ippSet functions
  * for all of the fields in these structures.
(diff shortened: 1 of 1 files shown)


zdohnal:
@tillkamppeter Hi Till, any update? I thought since cups-browsed connects only to local CUPS by default then we can use check for CUPS which is available during the build.

zdohnal:
@tillkamppeter Hi, any update?

tillkamppeter:
Note that the setting and removing of printer-is-shared is reqyuired to make a temporary queue permanent. If you skip it due to the fact that we have a remote CUPS queue and CUPS 2.2.x or newer, we must always proceed as we had an error during applying the printer-is-shared option, meaning that we remove the queue so that it gets re-created (the stuff in the following "`if (i <= 1) { }`").
Can you please modify the patch appropriately? Thanks.

zdohnal:
@tillkamppeter Till, are we both talking about the same part of code? It looks to me you mean the code where update_cups_queues() asks local CUPS for queue name, which cups-browsed has in remote_printers array, if CUPS returns a destination, then function checks if it is shared and if it is not, it makes queue permanent by setting printer-is-shared bit and it unset it right after. 

The pull request is about the part of code further down, where we already have permanent local queue created/modified by cups-browsed (about line 5342 and further) and we are trying to set printer-is-shared bit to indicate if our local queue should be shared, which is forbidden for queues pointing to remote CUPS queues since 2.1.X (I used 2.2 for just to be sure) and the request ends with harmless error - so IMHO doing request in such cases (queue pointing to remote cups queue+local CUPS > 2.2) just slows down cups-browsed and fills log with expected errors, which can confuse users.

Are we really talking about two different parts of code or do I just miss something?

tillkamppeter:
You're right, I confused it. Your patch is actually on the later place. So then it is OK this way. I will merge it.

zdohnal:
Thank you, Till! Have a nice day!

Issue #63, opened 2018-09-25
print a4 documents sensibly by default on command-line


archenemies:
When I print a PDF whose pages are smaller or larger than letter paper, I find that `lp` does a poor job of printing it. The printed page appears in the upper-left corner of the paper (medium) page, and the top half of the top line is cut off. Generally I print these documents with Evince and select "Fit to Printable Area" under "Page Handling", that does a better job - it scales up the page and centers it on the physical medium. I wonder if cups-filters can be configured to do the same thing as Evince is doing, it would be nice because I don't always remember to use Evince until after I've printed a document of many pages and found that it came out all wrong.

Here is an example document (one page, pdfinfo says "456 x 687.59 pts"):

[sample-8.pdf](https://github.com/OpenPrinting/cups-filters/files/2413783/sample-8.pdf)

archenemies:
Here's another example, it prints fine from Evince or Okular, but crashes my printer when I try to print it with 'lp':

[el-cerrito.pdf](https://github.com/OpenPrinting/cups-filters/files/2421842/el-cerrito.pdf)

archenemies:
I just discovered that there is already an option to 'lp' called 'fit-to-page', which makes at least the first example PDF print acceptably. (Does anyone read these issues?)

It's a pity that '-o fit-to-page' is not the default. Most users have to deal with a mix of A4 and letter documents, and it is difficult to tell which is which just by bringing it up in a PDF viewer. Should everyone with a letter printer have to remember to check whether a document is A4 before passing it to lp, and if so then specify '-o fit-to-page' manually? It would make so much more sense just to have this as the default.

On the other hand, I checked that passing "-o fit-to-page" when letter-printing a PDF document with letter-sized pages causes the whole page to shrink slightly. This is problematic when you are e.g. duplex-printing flashcards and need the front and back to line up; or if you are printing something where exact measurements are important. Not to mention that we would like to respect the margins set by the document.

Evince in fact gives the same output as normal 'lp' when you select "Page Handling > Page Scaling > None" in the Print dialog.

So I am renaming this to "print a4 documents sensibly by default on command-line" since I think that is a better summary of the problem.

It might be nice to see a default option like 'fit-to-page=ask' where if the document size is more than 1% different from the printer page size, and 'lp' is running on the terminal, then it will prompt the user for what to do...

tillkamppeter:
[...]

PR #92, opened 2019-02-18
Added print-scaling=fill functionality.

### LINKED ISSUE #65: Add a simple "fill page" scaling option to imagetoraster and pdftopdf (print-scaling=fill)
Imagetoraster, by default, scales the input image to fill the output page, while respecting the image's aspect ratio.

Unless the input and output aspect ratios are a near-perfect match, this often results in uneven borders/margins, or on full-bleed printers, unsightly white lines on one or more edges, and there is no simple way to avoid this without a-priori knowledge of the printer's native capabilities and using that in a pre-printing "filter" crop the image to the correct aspect ratio.

What I'd like to see is another scaling option that will scale the image's shorter dimension to match the printer's shorter dimension, then crop the excess off the longer dimension.  By default we would center the image in the page, but the "position" option could be used to change that.

This is the same idea behind Gutenprint's 'StpiShrinkOutput=Crop' option, but that feature is useless in a CUPS printing flow because imagetoraster generates a raster that already fits the aspect ratio.

On the surface this seems like it would be fairly straightforward to implement (given what imagetoraster already can accomplish), so if I bang this out, is it (in principle) something that would be accepted into cups-filters?

dheeraj135:
Related to #65 .

For this functionality, the input file has to cover the entire page. I first cropped input file according to the page's aspect ratio after that normal processing is done.
```imagetopdf``` and ```imagetoraster``` uses ```convert``` command(ImageMagick) to crop the input file.
```pdftopdf``` filter uses qpdf to crop the given pdf.
Is there any improvement that needs to be done in this PR?

### DIFF
--- a/filter/pdftopdf/pdftopdf.cc
+++ b/filter/pdftopdf/pdftopdf.cc
@@ -326,6 +326,12 @@ void getParameters(ppd_file_t *ppd,int num_options,cups_option_t *options,Proces
   // TODO?  pstops checks =="true", pdftops !is_false  ... pstops says: fitplot only for PS (i.e. not for PDF, cmp. cgpdftopdf)
   param.fitplot=(val)&&(!is_false(val));
 
+  if((val=cupsGetOption("print-scaling",num_options,options))!=NULL) {
+    if(!strcasecmp(val,"fill")) {
+      param.fillprint=true;
+    }
+  }
+
   if (ppd && (ppd->landscape < 0)) { // direction the printer rotates landscape (90 or -90)
     param.normal_landscape=ROT_270;
   } else {
@@ -1023,6 +1029,7 @@ int main(int argc,char **argv)
     param.nup.nupY=2;
     //param.nup.yalign=TOP;
     param.border=BorderType::NONE;
+    //param.fillprint = true;
     //param.mirror=true;
     //param.reverse=true;
     //param.numCopies=3;
--- a/filter/pdftopdf/pdftopdf.cc
+++ b/filter/pdftopdf/pdftopdf.cc
@@ -331,6 +331,21 @@ void getParameters(ppd_file_t *ppd,int num_options,cups_option_t *options,Proces
       param.fillprint=true;
     }
   }
+  else if((val = cupsGetOption("fill",num_options,options))!=0) {
+    if(!strcasecmp(val,"true")||!strcasecmp(val,"yes"))
+    {
+      param.fillprint = true;
+    }
+  }
+  /*
+   * crop-to-fit
+   */
+  if((val = cupsGetOption("crop-to-fit",num_options,options))!= NULL){
+    if(!strcasecmp(val,"true")||!strcasecmp(val,"yes"))
+    {
+      param.cropfit=1;
+    }
+  }
 
   if (ppd && (ppd->landscape < 0)) { // direction the printer rotates landscape (90 or -90)
     param.normal_landscape=ROT_270;
--- a/cupsfilters/image.c
+++ b/cupsfilters/image.c
@@ -807,3 +807,30 @@ get_tile(cups_image_t *img,		/* I - Image */
   return (ic->pixels + bpp * (y * CUPS_TILE_SIZE + x));
 }
 
+/*
+ * Crop a image.
+ * (posw,posh): Position of left corner
+ * (width,height): width and height of required image.
+ */
+cups_image_t* cupsImageCrop(cups_image_t* img,int posw,int posh,int width,int height)
+{
+  int image_width = cupsImageGetWidth(img);
+  cups_image_t* temp=calloc(sizeof(cups_image_t),1);
+  cups_ib_t *pixels=(cups_ib_t*)malloc(img->xsize*cupsImageGetDepth(img));
+  temp->cachefile = -1;
+  temp->max_ics = CUPS_TILE_MINIMUM;
+  temp->colorspace=img->colorspace;
+  temp->xppi = img->xppi;
+  temp->yppi = img->yppi;
+  temp->num_ics = 0;
+  temp->first =temp->last = NULL;
+  temp->tiles = NULL;
+  temp->xsize = width;
+  temp->ysize = height;
+  for(int i=posh;i<min(cupsImageGetHeight(img),posh+height);i++){
+    cupsImageGetRow(img,posw,i,min(width,image_width-posw),pixels);
+    _cupsImagePutRow(temp,0,i-posh,min(width,image_width-posw),pixels);
+  }
+  free(pixels);
+  return temp;
+}
--- a/cupsfilters/image.h
+++ b/cupsfilters/image.h
@@ -112,7 +112,8 @@ extern void		cupsImageWhiteToRGB(const cups_ib_t *in,
 			                    cups_ib_t *out, int count) _CUPS_API_1_2;
 extern void		cupsImageWhiteToWhite(const cups_ib_t *in,
 			                      cups_ib_t *out, int count) _CUPS_API_1_2;
-
+extern cups_image_t* 	cupsImageCrop(cups_image_t* img,int posw,
+					int posh,int width,int height);
 
 #  ifdef __cplusplus
 }
--- a/filter/imagetopdf.c
+++ b/filter/imagetopdf.c
@@ -693,7 +693,7 @@ main(int  argc,				/* I - Number of command-line arguments */
   int deviceReverse = 0;
   ppd_attr_t *attr;
   int pl,pr;
-
+  int fillprint = 0;  /* print-scaling = fill */
  /*
   * Make sure status messages are not buffered...
   */
@@ -810,6 +810,14 @@ main(int  argc,				/* I - Number of command-line arguments */
       }
   }
 
+  /*
+  *   print-scaling = fill functionality.
+  */
+  if((val = cupsGetOption("print-scaling",num_options,options)) !=0) {
+    if(!strcasecmp(val,"fill")) {
+        fillprint = 1;
+    }
+  }
 
   if ((val = cupsGetOption("OutputOrder",num_options,options)) != 0) {
     if (!strcasecmp(val, "Reverse")) {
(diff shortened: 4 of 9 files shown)

### CODE REVIEW

**Comment by tillkamppeter** [comment]
First, please take care of whitespace. Your patch contains many pure whitespace changes, like introducing new whitespace in originally empty lines. Also check the indentation of your code.

Second, with the "convert" utility of Image/GraphicsMagick you are introducing a new dependency. How did imagetopdf and imagetoraster work before? Or did they never do any cropping and/or scaling?
imagetopdf would perhaps even not need the new functionality if one takes care that after imagetopdf the print data goes through pdftopdf.

**Comment by tillkamppeter** [comment]
Did you also run all the test cases (from my comment in the issue) with your code?

**Comment by dheeraj135** [comment]
I was not able to find any cropping functionality in ```imagetoraster``` and ```imagetopdf``` filter. Other than cropping, we can stretch and let the image overflow with scaling and set the area to be printed. But cropping the input is cleaner and more reliable.

I tested the code with jpg, png and pdf files. I didn't understand "Test also with and without unprintable borders.". Maybe you can provide me with test files to test the code. 
For square orientation, we will have to specify a square paper size.
For other cases, the code seems to work fine.

I will remove white lines and correct the indentation.

**Comment by tillkamppeter** [comment]
I assume that the imageto... filters load the input image as some kind of bitmap into a data structure. You could go with loops through all pixels and for the crop only take the pixels which are inside the destination area to get the cropped image. AFAIR the rastertopwg filter of CUPS does such a thing (but I am not 100% sure).
"Test also with and without unprintable borders." means that you test with destination pages which have unprintable borders/margins (margin values > 0) and without borders/margins (margin values = 0, borderless, fullbleed).

**Comment by dheeraj135** [comment]
Ok. I will test the code with margins. 

I will check if cropping can be done without using convert.

**Comment by tillkamppeter** [comment]
For imagetopdf you can also change mime/cupsfilters.convs to let imagetopdf output application/pdf instead of application/vnd.cups-pdf and then pdftopdf is used and will do the scaling. But make sure that any existing page management option is always applied only once and not twice.

**Comment by dheeraj135** [comment]
@tillkamppeter 
I used the code with page-{top, bottom, left, right} for margins and it was giving expected output. 
Code no longer depends on convert. A new function is added to cupsfilters/image.h for cropping the image. Both ```imagetoraster``` and ```imagetopdf``` uses this function for cropping.
It is better to make the required changes in ```imagetopdf``` rather than using it with ```pdftopdf``` because then extra overhead will be incurred.

**Comment by tillkamppeter** [comment]
Grsat, thank you very much.
Does it now cover all [these](https://github.com/OpenPrinting/cups-filters/issues/65#issuecomment-465122873) cases?

**Comment by tillkamppeter** [comment]
Could you also check the specs of IPP for whether there are options/settings defined which are equivalent with the new cases you added (`fill`, `crop-to-fit`) so that filters will also understand the standards-conforming IPP attributes?

**Comment by dheeraj135** [comment]
@tillkamppeter I still have to add ```crop-to-fit``` option. I am working on it. As far as *fill* is considered, all cases are covered, but I didn't test them on a real printer. It would be great if you could also test *fill* option.

File: filter/pdftopdf/pdftopdf.cc [hunk 1]
```diff
@@ -750,6 +771,15 @@ bool checkFeature(const char *feature, int num_options, cups_option_t *options)
...
   }
+  /*
+   *  To make sure fillprint or cropfit is not applied twice.
+   */
+  char* content_type = getenv("CONTENT_TYPE");
+  if(!strncasecmp(content_type,"image",5))
+  {
+    param.fillprint =0;
+    param.cropfit =0;
+  }
```
dheeraj135: @tillkamppeter Is this check sufficient to make sure fillprint and cropfit are not applied again in ```pdftopdf``` filter?

dheeraj135:
```crop-to-fit``` feature has been added. ```crop-to-fit``` is not defined in PWG standards. 
```-o fill``` option can be used to use ```print-scaling=fill``` feature. 
I tested the code and here are the results: [Google Drive](https://drive.google.com/open?id=1bzbrLUyRS62t18qfJk3omOC3jRQPSV6O).

**Comment by dheeraj135** [comment]
@tillkamppeter any updates on this PR?

File: filter/pdftopdf/pdftopdf.cc (hunk 1 above)
Comment (tillkamppeter) [reply filter/pdftopdf/pdftopdf.cc]: You are checking via
```
char* content_type = getenv("CONTENT_TYPE");
￼if(!strncasecmp(content_type,"image",5)) ...
```
meaning that if the file type in which the job got sent is any MIME type starting with "image". This would also include "image/urf", "image/pwg-raster", ... which are formats not treated by the "`imageto...`" filters and so they need the "fit-to-page", "fill", "crop-to-fit" option be applied by "`pdftopdf`".
If it comes to the use of "`imagetoraster`", "`pdftopdf`" is never called. If "`imagetopdf`" is used we must decide. If we use "`imagetopdf`" we need to make it mark its PDF output, so that "`pdftopdf`" knows that there was "`imagetopdf`" before and "fit-to-page", "fill", "crop-to-fit" do not need to be done any more. The marking can be done the same way as "`pdftopdf`" does to tell "`pdftops`" about Copies and Collate, by adding a comment. "`imagetopdf`" could add "%%from imagetopdf" and "`pdftopdf`" checks whether the input stream contains it, then does not apply the scaling/crop options any more.
Please try it this way and test.

**Comment by tillkamppeter** [comment]
Please see my answer about avoiding duplicate applications of the options above.

File: filter/pdftopdf/pdftopdf.cc (hunk 1 above)
Comment (dheeraj135) [reply filter/pdftopdf/pdftopdf.cc]: Hi Till,
I noticed that imagetopdf gives ```application/vnd.cups-pdf``` as output format. 
```
 ~/g/c/mime> grep -R "imagetopdf" *
cupsfilters.convs:image/gif		application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs:image/png		application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs:image/jpeg		application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs:image/tiff		application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs:image/x-bitmap		application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs:image/x-photocd		application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs:image/x-portable-anymap	application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs:image/x-portable-bitmap	application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs:image/x-portable-graymap application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs:image/x-portable-pixmap	application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs:image/x-sgi-rgb		application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs:image/x-xbitmap		application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs:image/x-xpixmap		application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs:image/x-xwindowdump	application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs:image/x-sun-raster	application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs.in:image/gif		application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs.in:image/png		application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs.in:image/jpeg		application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs.in:image/tiff		application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs.in:image/x-bitmap		application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs.in:image/x-photocd		application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs.in:image/x-portable-anymap	application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs.in:image/x-portable-bitmap	application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs.in:image/x-portable-graymap application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs.in:image/x-portable-pixmap	application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs.in:image/x-sgi-rgb		application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs.in:image/x-xbitmap		application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs.in:image/x-xpixmap		application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs.in:image/x-xwindowdump	application/vnd.cups-pdf		65	imagetopdf
cupsfilters.convs.in:image/x-sun-raster	application/vnd.cups-pdf		65	imagetopdf

```
and pdftopdf converts ```application/pdf``` to ```application/vnd.cups-pdf```: 
```
 ~/g/c/mime> grep -R "pdftopdf" *
cupsfilters.convs:# that pdftopdf is doing the page management on PDF data and the
cupsfilters.convs:application/pdf		application/vnd.cups-pdf		66	pdftopdf
cupsfilters.convs.in:# that pdftopdf is doing the page management on PDF data and the
cupsfilters.convs.in:application/pdf		application/vnd.cups-pdf		66	pdftopdf
```
So, imagetopdf and pdftopdf will not work on same input. Am I missing something?

tillkamppeter: OK, you are right. Then this problem will not occur.
Only make sure that when it comes to an `imagetopdf`-based filter chain that the common page management options (if they apply, does `imagetopdf` accept any multi-page format as input ?) work.

dheeraj135: Which functionalities do I need to test? I don't think ```imagetopdf```  accepts multi-page format. One known issue is, If we specify orientation with ```crop-to-fit``` we get wrong results in ```imagetopdf``` and ```imagetoraster```. ```crop-to-fit``` is not intended to be used with orientation. Is there any way to ignore orientation if ```crop-to-fit``` is specified?

tillkamppeter: I see that multi-page output is only possible if you scale the image larger than the output page size to allow for a multi-page poster. With this duplex, reverse order, collate, number-up, and selected pages make sense, but applying any option which you have added your force the filter to take a single input page and produce a single output page from it. So page management options do not make sense.
crop-to-fit does only make sense for input which has an absolute size defined, like PDF or PostScript. On images, like photos, which do not have an absolute size defined (only being a bitmap) crop-to-fit does not make sense and should be ignored (fall back to default, fit-to-page).
Orientation with crop-to-fit is handled as without crop-to-fit. If not forced by orientation-requested the page is rotated if its orientation does not match the output page orientation (portrait on most printers) and after that crop-to-fit is applied.
So please make sure that it behaves this way and if so, tell me, so that I can merge the pull request.

dheeraj135: I have made the required changes. PR is ready from my side. Please go through the PR once more.

tillkamppeter:
Did you implement crop-to-fit now as I tolde above:
> Orientation with crop-to-fit is handled as without crop-to-fit. If not forced by orientation-requested the page is rotated if its orientation does not match the output page orientation (portrait on most printers) and after that crop-to-fit is applied.

So crop-to-fit goes like default mode without any scaling, starting to rotate the page if it is not the same orientation as the destination page (for example in most printers paper is supplied in portrait orientation, so an incoming landscape-oriented page gets rotated, but after that instead of requesting the input page's page size it prints on the destination page size (as requested via the "PageSize" or "media" option or PPD default setting), cropping it to fit. This should solve issue #63.

dheeraj135:
For images, I am checking if more than 50% of height or width is to be cropped in ```crop-to-fit```, then we ignore it. Orientation is working fine for all the filters. If orientation is not specified, I am rotating images in ```imageto...``` filters to cover the maximum area of the page but in ```pdftopdf```, I am not doing this because pdf is supposed to have a standard size.

dheeraj135:
Please go through PR once more. I think it can be merged now. It is ready from my side.

tillkamppeter:
What do you mean with
> ... but in pdftopdf, I am not doing this because pdf is supposed to have a standard size.

Note that if the input page is Letter in Landscape and the printer uses Letter paper (inserted in Portrait orientation, short edge entering first) then the pdftopdf filter rotates the page so that the content gets correctly onto the output sheet.
Now if we have again an input page in Letter size with Landscape orientation and the printer uses A4, also in Portrait orientation, short edge first, and we use crop-to-fit, we will also need the rotation before we apply  the crop-to-fit, so that the content lands correctly on the output page.
Will this happen with your current code?

dheeraj135:
Input file: [margin-test.pdf](https://github.com/OpenPrinting/cups-filters/files/2936187/margin-test.pdf)
**Default**
Command: ```./lpr -P t5 /home/dh/Downloads/margin-test.pdf```
Output file: [18-margin-test_pdf.pdf](https://github.com/OpenPrinting/cups-filters/files/2936243/18-margin-test_pdf.pdf)
**Fitplot**
Command: ```./lpr -P t5 /home/dh/Downloads/margin-test.pdf -o media=A4 -o fitplot```
Output file: [19-margin-test_pdf.pdf](https://github.com/OpenPrinting/cups-filters/files/2936255/19-margin-test_pdf.pdf)
**Crop-to-fit**
Command: ```./lpr -P t5 /home/dh/Downloads/margin-test.pdf -o media=A4 -o crop-to-fit```
Output file: [16-margin-test_pdf.pdf](https://github.com/OpenPrinting/cups-filters/files/2936197/16-margin-test_pdf.pdf)
**print-scaling=fill**
Command: ```./lpr -P t5 /home/dh/Downloads/margin-test.pdf -o media=A4 -o fill```
Output file: [17-margin-test_pdf.pdf](https://github.com/OpenPrinting/cups-filters/files/2936209/17-margin-test_pdf.pdf)
I hope this clarifies everything.

tillkamppeter:
Tests look all correct for me. Thanks.

PR #95, opened 2019-02-28
Replace cupsTempFd() with tmpfile()

ppawliczek:


tillkamppeter:
Why do you want to replace cupsTempFd() with tmpfile()?

tillkamppeter:
Thank you very much for your contribution.
Unfortunately, we will not merge it into our code base, as we decided to stay with the cupsTempFd() function here for the following reasons:
- We need not only to have a temporary file opened but we also need to have access to its file name. tmpfile() only gives us a file descriptor so that we can read and write the file, but not the file name. The separation of file name/file descriptor creation and actual file creation should not be a problem as the file name is not predictable.
- We need to control with environment variables in which directory the temporary file is created. As cups-filters will be used in several sand-boxed packages, like CUPS and also Printer Applications we cannot use tmpfile() where it is not assured that it follows temporary variable for the directory where it drops the file.
- The method of how it determines the file name does perhaps produce less different names in a short time but still enough for our use cases (See also apple/cups#3382).

PR #98, opened 2019-03-07
Set orientation for crop-to-fit and fill in pdftopdf.

dheeraj135:
Bug fix for #92. 
When the input format is other than "pdf"(like ps) and ```pdftopdf``` filter chain is being used then ```pdftopdf``` is unable to automatically set the orientation for crop-to-fit and fill options. When orientation is not set, the is_landscape function is used to check orientation of the input file.


tillkamppeter:
Thanks for the fix. I have applied it now.

PR #103, opened 2019-04-02
cups-browsed: Fix broken trailing space removal on "NickName".

olifre:
The code tried to remove trailing spaces, but if there was
at least one trailing space, the last two characters of the
line were removed. This caused removal of the quotation mark
in case exactly one trailing space was present, breaking
the PPD and subsequently causing CUPS to mess up all queues
(even local ones).

Signed-off-by: Oliver Freyermuth <o.freyermuth@googlemail.com>


tillkamppeter:
Thank you very much for the patch. I have applied it now.

PR #105, opened 2019-04-07
Correct comment

albert-github:
Doxygen throws a warning: warning: the name `urf_decode.cpp' supplied as the second argument in the \file statement is not an input file
this is corrected. The information should be correct regardless whether or not it is used.


tillkamppeter:
Thank you for the patch. I have merged it now.

PR #107, opened 2019-04-09
Switched {image→graphics}magick

### LINKED ISSUE #106: Switch from imagemagick to graphicksmagick (or another library)
Imagemagick is problematic since it hijacks a lot of command names (convert, import, …) and has quite a few security problems. There’s more here: http://www.graphicsmagick.org/

flying-sheep:
This looks correct, but I don’t know how to test it properly. Do you have any hints?

Fixes #106


tillkamppeter:
This patch is completely within the Braille embosser support which is contributed by @sthibaul, so best is if @sthibaul tests this.
@sthibaul, could you please test this patch which convertsthe Braille embosser support from using ImageMagick to GraphicsMagick? Thanks.

sthibaul:
graphicsmagick does not support the .ubrl and .brf formats, so this will just not work at all.

tillkamppeter:
@sthibaul, does this mean that we are forced to use ImageMagick here? Is this the only free software platform which exists that supports these formats?

sthibaul:
AFAIK it's the only software that supports them, yes.

tillkamppeter:
@sthibaul, OK, thanks for the info. I would recommend you to make a feature request to GraphicsMagick to overtake the support for `.ubrl` and `.brf` from ImageMagick.
@flying-sheep, thank you very much for the patch, but as it breaks the functionality of our Braille embosser support, we cannot apply it.

flying-sheep:
of course, makes sense! damn, never heard of those formats, it might indeed be hard to find a replacement.

sthibaul:
I have submitted a patch to graphicsmagick:

https://sourceforge.net/p/graphicsmagick/patches/59/

PR #118, opened 2019-05-11
Add support for print-scaling option.

### LINKED ISSUE #108: Support "print-scaling" option
The "print-scaling" option is the standard IPP Everywhere option for controlling the scaling of the output; currently defined values are:

- `auto`: if the “ipp-attribute-fidelity” attribute is true or the document is larger than the requested media, scale the document using the 'fit' method if the margins are non- zero, otherwise scale using the 'fill' method. If the “ipp-attribute-fidelity” attribute is false or unspecified and the document is smaller than the requested media, scale using the 'none' method.
- `auto-fit`: if the “ipp-attribute-fidelity” attribute is true or the document is larger than the requested media, scale the document using the ‘fit’ method. Otherwise, scale using the ‘none’ method.
- `fill`: scale the document to fill the requested media size, preserving the aspect ratio of the document data but potentially cropping portions of the document.
- `fit`: scale the document to fit the printable area of the requested media size, preserving the aspect ratio of the document data without cropping the document.
- `none`: do not scale the document to fit the requested media size. If the document is larger than the requested media, center and clip the resulting output. If the document is smaller than the requested media, center the resulting output.

This attribute is defined in [PWG 5100.16: IPP Transaction-Based Printing Extensions](https://ftp.pwg.org/pub/pwg/candidates/cs-ipptrans10-20131108-5100.16.pdf).

dheeraj135:
Support for the print-scaling feature has been added to imagetoraster, imagetopdf and pdftopdf filters.
Fixes #108.

### DIFF
--- a/filter/imagetopdf.c
+++ b/filter/imagetopdf.c
@@ -811,30 +811,6 @@ main(int  argc,				/* I - Number of command-line arguments */
       }
   }
 
-  /*
-  *   print-scaling = fill functionality.
-  */
-  if((val = cupsGetOption("print-scaling",num_options,options)) !=0) {
-    if(!strcasecmp(val,"fill")) {
-        fillprint = 1;
-    }
-  }
-  else if((val = cupsGetOption("fill",num_options,options))!=0) {
-    if(!strcasecmp(val,"true")||!strcasecmp(val,"yes"))
-    {
-      fillprint = 1;
-    }
-  }
-  /*
-   * crop-to-fit
-   */
-  if((val = cupsGetOption("crop-to-fit",num_options,options))!= NULL){
-    if(!strcasecmp(val,"true")||!strcasecmp(val,"yes"))
-    {
-      cropfit=1;
-    }
-  }
-
   if ((val = cupsGetOption("OutputOrder",num_options,options)) != 0) {
     if (!strcasecmp(val, "Reverse")) {
       Reverse = 1;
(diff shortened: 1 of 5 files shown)


filter/imagetopdf.c
```diff
@@ -997,6 +958,113 @@ main(int  argc,				/* I - Number of command-line arguments */
   colorspace = ColorDevice ? CUPS_IMAGE_RGB_CMYK : CUPS_IMAGE_WHITE;
 
   img = cupsImageOpen(filename, colorspace, CUPS_IMAGE_WHITE, sat, hue, NULL);
+  if(img!=NULL){
+
+  int margin_defined = 0;
+  int fidelity = 0;
+  int document_large = 0;
+
+  if(ppd->custom_margins[0]||ppd->custom_margins[1]||ppd->custom_margins[2]||ppd->custom_margins[3])
```
dheeraj135: @tillkamppeter Is this the right way to check for margins?

tillkamppeter: I think not, AFAIR the custom_margins[] array is for the margins for a custom page size. If as output page size one of the list of the PageSize option is chosen, the margins assigned to this size (from the ImageableArea lines in the PPD, should be also somewhere in CUPS' ppd data structure) are the ones you are looking for.

dheeraj135: Thanks, @tillkamppeter. I have updated the PR.

tillkamppeter:
Dheeraj, thank you verymuch for your contribution. I have merged it now so that it will make part of the 1.23.0 release of cups-filters.

PR #121, opened 2019-05-20
Corrected Code Styling

deepak0405:
- Corrected code styling for "for" and "if"
- Handling dup() status in the implicitclass backend

### DIFF
--- a/backend/implicitclass.c
+++ b/backend/implicitclass.c
@@ -138,7 +138,7 @@ main(int  argc,				/* I - Number of command-line args */
   int     bytes;      /* Bytes copied */
   char uri[HTTP_MAX_URI];
   char    *argv_nt[7];
-  int     outbuflen,filefd,exit_status;
+  int     outbuflen,filefd,exit_status,dup_status;
   static const char *pattrs[] =
                 {
                   "printer-defaults"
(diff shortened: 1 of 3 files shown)


tillkamppeter:
Thank you for the fixes, but I still have one question:
Did you really switch to case-insensitive search/comparison EVERYWHERE where option names or choice names are compared/searched?

deepak0405:
1.  Most of the part of the code for the option is taken from the ppdgenerator.c. At some places I used 
    strcmp as it was used in the ppdgenerator file. like

  ```
    if  (!have_bi_level && (!strcasecmp(keyword, "black_1") ||
			     !strcmp(keyword, "bi-level") ||
			     !strcmp(keyword, "process-bi-level"))) {
  ```

  ```
   !strcmp(side, "two-sided-long-edge")) {
  ```
The ppdgenerator.c is not comparing some options case-sensitively, and I did the same for those options in cups-browsed.

 2.  The function supports_job_attributes_requested() compares the job setting with the options supported by the printer. The comparisons here are case-insensitive as :
```
      if (!cupsArrayFind(media_type_supported, (void *)str) &&
	  strcasecmp(str,AUTO_OPTION)) {
```
We are finding the job-option requested in the array of job-options of the printer. And in the constructor of the array we used strcasecmp.
```
  if ((media_data = cupsArrayNew3((cups_array_func_t)strcasecmp, NULL, NULL, 0,
				  (cups_acopy_func_t)strdup,
				  (cups_afree_func_t)free)) == NULL)
```

Issue #9, opened 2017-12-13
pdftoraster uses non-public/internal APIs of Poppler


tillkamppeter:
@sahilarora535 investigated the code of pdftoraster and writes:

> After going through the code, I found that a lot of poppler APIs are being used in [pdftoraster.cxx](https://github.com/OpenPrinting/cups-filters/blob/b72c0bf80f21085521508900130ec553fae089b7/filter/pdftoraster.cxx#L39) (please open the link which redirects to the line number). When I built cups-filters for the first time, I distinctively remember I had to install poppler as well since it is a dependency for cups-filters. However, even on installing, files such as “goo/gmem.h” (which are included in the filter) were not found, hence I had to make poppler from the source. Even that did not work. Hence, I had to use a flag which configuring poppler so that these files come along, and the command was *|./configure --enable-xpdf-headers|* to configure poppler. I guess these are the unstable APIs we have been talking about. My first impression of the code base is that pdftoraster has a lot of uses of poppler’s unstable APIs, and this is what it is. Please correct me if I am wrong.

We need an existing Poppler output device for raster output ot create one if there is no suitable one. Poppler has a `pdftoppm` utility which generates the genric PPM raster format, so Poppler's output device used by this utility can probably somehow made use of. I do not know the PPM and similar raster graphics formats, for example whether all color spaces available in PWG/CUPS Raster are supported.
Here we have different possibilities depending on what the output formats of `pdftoppm` are capable of:

1. A pdftoraster filter which calls `pdftoppm` on the command line and converts the output into PWG/CUPS Raster.
2. A pdftoraster filter which uses the Poppler output device which the `pdftoppm` utility uses, using only public APIs of Poppler and it wraps the output bitmap streams into valid PWG/CUPS Raster.
3. Either we extend the PPM/raster output device of Poppler to also support PWG/CUPS Raster or we create a new Poppler output device for PWG/CUPS Raster based on Poppler's existing PPM/raster output device. The changes on Poppler we have to submit upstream. Then we do a pdftoraster filter which uses this device.

In any case the new pdftoraster filter should not loose any features, especially support for all color spaces and color management as the current one has. 
Note: The same is probably the case for pdftoijs, pdftoopvp, and bannertopdf, but the former two we are deprecating curremtly (they are not installed by default any more) and the latter we will convert to use QPDF (Issue #8).

sahilarora535:
[...]

PR #125, opened 2019-06-05
Integration of Stable API of poppler

tanmayanand44:
Work done as part of GSoC-19. Fixes issue #9 
I have completely removed all unstable API of poppler and integrated the stable API of poppler-cpp.
I will be adding some test reports as well in the comments.
Currently supported colorpspaces: black/gray,sgray,rgb,adobergb,cmyk and srgb
You may test using these examples: https://ftp.pwg.org/pub/pwg/ipp/examples/originals/


tanmayanand44:
All the test reports are available at https://github.com/tanmayanand44/pdftoraster-gsoc19-test_reports
Note: it is a 400mb folder

PR #139, opened 2019-08-12
Fix segfault when running foomatic-rip by hand and env PRINTER is mis…

zdohnal:
…sing

Hi,

I was debugging foomatic-rip filter by hand and I ran into segfault which is caused by my pull request :( . I hope this should finally fix it...

Would you mind merging it into the project?


tillkamppeter:
Merged, thank you very much for your patch.

PR #154, opened 2019-09-16
install: support old ln versions without the -r option

casantos:
Oldish enterprise-class distributions have too old versions of
coreutils, with ln not supporting -r.

So we fake it.

ln -r would create minimalist relative paths, but they are not
trivial to generate. Instead, we always create paths relative to the
root, i.e.:

    ln -s -r /usr/bin/foo /usr/sbin/foo

would create:  /usr/sbin/foo -> ../bin/foo
while we do :  /usr/sbin/foo -> ../../usr/bin/foo

Signed-off-by: Olivier Schonken <olivier.schonken@gmail.com>


tillkamppeter:
There should be some switch in ./configure to detect whether "`ln -f`" is availableand then be used "`ln -f`" or the alternative/workaround.
Does your workaround work for any combination of CUPS and cups-filters being in `/usr` or `/usr/local`?

casantos:
> There should be some switch in ./configure to detect whether "`ln -r`" is availableand then be used "`ln -r`" or the alternative/workaround.

I agree but at the moment I'm focusing on the functionality.

> Does your workaround work for any combination of CUPS and cups-filters being in `/usr` or `/usr/local`?

I just submitted an update that uses a shell script to mimic ln -r. It meets that requirement.

casantos:
New version, with logic to detect if ln supports -r.

Makefile.am
```diff
@@ -988,14 +988,14 @@ install-exec-hook:
...
 endif
 if ENABLE_DRIVERLESS
-	$(LN_S) -r -f $(DESTDIR)$(pkgppdgendir)/driverless $(DESTDIR)$(bindir)
-	$(LN_S) -r -f $(DESTDIR)$(pkgppdgendir)/driverless $(DESTDIR)$(pkgbackenddir)
+	$(LN_SRF) $(DESTDIR)$(pkgppdgendir)/driverless $(DESTDIR)$(bindir)
+	$(LN_SRF) $(DESTDIR)$(pkgppdgendir)/driverless $(DESTDIR)$(pkgbackenddir)
 endif
 if ENABLE_BRAILLE
-	$(LN_S) -f imagetobrf $(DESTDIR)$(pkgfilterdir)/imagetoubrl
+	$(LN_S) -f imagetobrf $(DESTDIR)$(pkgfilterdir)/imagetoubrlR
```
tillkamppeter: What is this change good for? Looks like a typo.

Makefile.am
```diff
@@ -1009,7 +1009,7 @@ endif
 install-data-hook:
 if RCLINKS
 	for level in $(RCLEVELS); do \
-		$(INSTALL) -d -m 755 $(DESTDIR)$(INITDIR)/rc$${level}.d; \
+		$(INSTALL) -d -m 755 $(DESTDIR)$(INITDIR)/rc$${level}.d; \R
```
tillkamppeter: What is this change good for? Looks like a typo.

tillkamppeter:
Thank you for the update. I have added 2 comments to your commit. There seem to be random typos in it.

tillkamppeter:
Fixes #49.

casantos:
On Wed, Sep 18, 2019 at 5:01 AM Till Kamppeter <notifications@github.com> wrote:
>
> Thank you for the update. I have added 2 comments to your commit. There seem to be random typos in it.

Fixed

-- 
Carlos Santos <unixmania@gmail.com>

tillkamppeter:
Thank you very much for your contribution. I have merged it now.

Issue #152, opened 2019-09-05
cups-browsed: Inappropriate PPD file generated


debiantriage:
I am using cups-browsed 1.25.4-2 on Debian. The generated PPD for an ENVY4500 has the line

> *cupsFilter2: "application/vnd.cups-pdf application/pdf 0 -"

The TXT record has

> pdl=application/vnd.hp-PCL,image/jpeg,application/PCLm,image/urf

Consequently, printing to the device will fail.

This is Debian Bug #939476:

[https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=939476](url)

Regards,

Brian.

tillkamppeter:
Does your print queue point directly to a printer (device URI is `ipp://...` or `ipps://...` or to a remote CUPS queue for the printer (device URI is `implicitclass://...`)?

debiantriage:
`lpstat -t` gives me

> implicitclass://ENVY4500/

Is that what you want?

-- 
Brian.

tillkamppeter:
Yes, this means that the queue with the PDF line added to the PPD file is not pointing to the printer directly, but to a CUPS server which hosts this printer. This means that your queue has to send PDF to the remote CUPS server as that one has the driver for this printer set up. That server converts the PDF into one of the printer's PDLs (PCL or Apple Raster, depends on the PPD which the queue on the remote server uses). So this PPD is correct.
If the printer does not print, we need debug log output of cups-browsed on your client and of CUPS on your remote server.

debiantriage:
There is a single computer on this network with a CUPS server and sharing its queues. I put _BrowseLocalProtocols none_ in its _cupsd.conf_ and restarted cupsd.  _avahi-browse -art_ shows the ENVY4500 as the only printing device on the network. The entry disappeared when the device was switched off.

I have a Debian unstable installation on another partition. It has not been updated and has cups-browsed 1.21.6-3. _lpstat -t_ indeed shows the device UR as _ipp://..._. However, I am convinced that the printer is being directly contacted on both installations.

I have attached debug log outputs for cups-browsed 1.21.6-3 and 1.25.4-2 and note that both Remote printer overview sections have an _ipp://..._ URI.

An everywhere queue is correctly set up without a *cupsFilter2 line for PDF (as is a non-auto driverless queue).

Regards,

Brian.
[debug_log-1.21.6-3.txt](https://github.com/OpenPrinting/cups-filters/files/3583971/debug_log-1.21.6-3.txt)
[...]

PR #159, opened 2019-09-23
implicitclass: Set the value of dest_host

deepak0405:
The value of dest_host was not set, because of which `implicitclass` was not handling the `NO_DEST_FOUND` message. Corrected the same in this PR.

Also the cups is setting the value of `cups-browsed-dest-printer` in this case, which should not be sent to the `ipp` backend, so removing that from arguments.

The uri status `HTTP_URI_STATUS_UNKNOWN_SCHEME` is a warning and should be ignored.


tillkamppeter:
Could you clean up your pull request somewhat? Your intended fixes are two changes (first and last commit) in implicitclass.c. The addition of ipp.patch does not make sense at all. Please apply your intended patches to a fresh copy of the original repository. Thanks.

tillkamppeter:
Fixes #152.

deepak0405:
Hey @tillkamppeter, I don't know why git is showing changes in `cups-browsed.c`, I checked the file which is present online, it is not having those changes which the diff is showing.

deepak0405:
@tillkamppeter Corrected the PR.

PR #168, opened 2019-10-29
Epson ET-7700 does not support image/pwg-raster

zdohnal:
Hi,

I have report in Fedora  - https://bugzilla.redhat.com/show_bug.cgi?id=1756726 - which is about Epson ET-7700 reports bad document format when image/pwg-raster format is used.
It seems the model does not support pwg-raster, but image/urf format works fine.

The patch is about crossing out that specific model from pwg-raster document format, so output ppd does not have the line for pwg-raster.

### DIFF
--- a/cupsfilters/ppdgenerator.c
+++ b/cupsfilters/ppdgenerator.c
@@ -1868,7 +1868,9 @@ ppdCreateFromIPP2(char         *buffer,          /* I - Filename buffer */
     formatfound = 1;
     is_pdf = 1;
   }
-  if (cupsArrayFind(pdl_list, "image/pwg-raster")) {
+  /* Epson ET-7700 reports pwg-raster support in get-printer-attributes response, but it does
+     not support pwg-raster in reality - all jobs ends with */
+  if (cupsArrayFind(pdl_list, "image/pwg-raster") && strncmp(model, "ET-7700 Series", 14)) {
     if ((attr = ippFindAttribute(response,
 				 "pwg-raster-document-resolution-supported",
 				 IPP_TAG_RESOLUTION)) != NULL) {


tillkamppeter:
Thank you very much for your patch.
This patch is at best a temporary solution, but not what should be permanent part of cups-filters. It works around a firmware bug of one single printer model (the bug being either the printer reporting it does PWG Raster but PWG Raster support not implemented in the printer, or the printer's PWG Raster support simply being buggy).
As there can be more printers with firmware bugs it would be more useful to have a quirk rule file, like CUPS has for the USB communication with USB printers:
```
/usr/share/cups/usb/*.quirks
```
cups-filters should also get quirk rule files, one for IPP firmware bugs, telling for each buggy printer model which advertised printer features are not usable due to a bug.
Quirk rule files give a nice overview of firmware bugs and workarounds, and they are text files so a sys admin, an experienced user, or a reporter of a bug guided by a developer/package maintainer can add rules and try out what makes his printer working again. In case of success one only needs to report an issue or pull request here to get the new rule included.

zdohnal:
Ok, that sounds good, I'll rewrite it into the way similar to CUPS quirks.

Do you have any suggestion about structure of quirk file structure?

I would propose the scheme "manufacturer  - model - quirk", so for our case would it be:

Epson - ET-7700 Series - no-pwg-raster

The quirks will be loaded at the start of ppdCreateFromIPP2() function, and they will be checked only when there will be image/pwg-raster in document-format-supported attribute.
I would not go into defining other feature names or putting the quirk check into other parts of code unless we find other firmware bug.

tillkamppeter:
I suggest something like this for the quirk file:
```
make model-regexp wrong-attr attr=value attr=value1,value2,... ...
make model-regexp get-printer-attributes arg1 arg2 ...
```
The first form is to tell about wrong attributes in the response to the get-printer-attributes request, in our case it would be `image/pwg-raster` under `document-format-supported` so the quirk rule line would be
```
"Epson" "ET-7700 Series" wrong-attr document-format-supported=image/pwg-raster
```
before generating the PPD these would be "substracted" from the attribute in the get-printer-attributes IPP response, so in this case the PPD would be built based on the other listed formats.
This syntax would also allow to describe broken arguments in more than one attribute. Imagine our printer reports to do 1200 dpi and 2400 dpi but trying to print in these resolutions gives broken output. So you extend the line to
```
"Epson" "ET-7700 Series" wrong-attr document-format-supported=image/pwg-raster printer-resolution-supported=1200dpi,2400dpi
```
If a printer misbehaves (or even crashes) on the standard get-printer-attributes IPP request with the arguments "all" and "media-col-database" one could use the second form to let the get-printer-attributes IPP request be done with other arguments. For example
```
"Canon" ".*" get-printer-attributes all
```
would let the request be done with only the "all" argument on all Canon printers.
More than one line should be allowed for one and the same printer, to allow both forms for the same printer if needed and also for example to tell about a wrong argument on a series of printers of one manufacturer and another wrong argument on one individual printer in the series. So after a line with matches manufacturer and model the rest of the quirk file should be read to find further matches.
I think this should cover most possible firmware bugs when requesting printer capabilities via the get-printer-attributes IPP request.

tillkamppeter:
Also the Epson WF-2630 shows this problem. See issue #171.

tillkamppeter:
The third printer known to show this problem is the Canon Pixma Pro 10, according to apple/cups#5238.

tillkamppeter:
As the only known quirk in driverless printing is PWG Raster support reported by printers which are not able to actually print it, we follow CUPS as in apple/cups#5238 for now letting the generated PPD make printing done in Apple Raster when both PWG Raster and Apple Raster are supported.
This way we can get away without quirk rule file for now, as this would also be somewhat against the principles of driverless printing. Therefore I am closing this issue for now.
@zdohnal, thank you anyway for offering to add quirk rule support.

tillkamppeter:
By the way, the fix is commit 80f87b0a.

zdohnal:
Ok, thank you for the info.

fnevgeny:
FYI, same issue with Epson L3160 (perhaps common for all EcoTank's?). BTW, the CUPS debug showed

`D [04/Jan/2020:21:34:07 +0200] [Job 91] Send-Document: client-error-document-format-error (client-error-document-format-error)`

which, ideally, should have been an error message.

tillkamppeter:
@fnevgeny, did you try the fix of the bug presented here (or a newer cups-filters version which contains it)?

fnevgeny:
Yes, sure, removing the "pwg-raster ..." string helped; sorry for not being clear about it.

fnevgeny:
Sorry again - if you mean the patch of cups-filters, no I didn't do it. I just fixed the ppd file.

tillkamppeter:
@fnevgeny, your edit on the PPD file has the same effect as the cups-filters fix, so all is OK for you and with the new cups-filters you do not need to edit your PPD any more.

PR #172, opened 2019-11-07
cups-browsed: Fix unset location check to use DNS-SD field

timor:
In `examine_discovered_printer_record`, the `location` parameter is ensured not to be `NULL`.  However, when checking whether to update the location from the DNS-SD TXT record note field, it is checked for being a `NULL` pointer.  Since all invocations seem to default to the empty string instead of `NULL`, the check is adjusted accordingly.


tillkamppeter:
Thank you very much for your patch. I have merged it now.

PR #179, opened 2019-11-27
Fix leaks of ipp_t struct and load balancing on the servers

zdohnal:
Hi,

we had an important issue in RHEL/Fedora about cups-browsed not closing http sockets to local CUPS in on_job_state(), which led into CUPS denial of service (MaxClient of cupsd was reached). I checked the latest upstream code and found out following issues:

1) httpConnectEncryptShortTimeout was removed from on_job_state(), but LoadBalancing on servers still needs http connection to remote server through http variable - currently it stays NULL and connection to remote server does not work. The patch puts it back, sets the correct 'if' and free the socket.
2) missing several cleanup of 'response' struct

The pull request fixes it, would you mind adding it to the project?

### DIFF
--- a/utils/cups-browsed.c
+++ b/utils/cups-browsed.c
@@ -3029,6 +3029,7 @@ int supports_job_attributes_requested(const gchar* printer, int printer_index,
   int                   i,count,side_found,orien_req,orien,
                         orien_found;
   cups_array_t          *sizes;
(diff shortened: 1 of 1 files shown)


tillkamppeter:
Thank you for the patch. I have applied it now.

PR #180, opened 2019-11-27
Do not set printer-is-shared for remote queue vol. 2

zdohnal:
Hi Till,

this is basically the same commit as #91 , but for the code you mentioned at first time - breaking from loop if the queue is remote should do the trick (cups-browsed at least creates correct print queue and printing is ok).

Would you mind adding it to the project too and prevent bogus errors?


tillkamppeter:
Thank you very much for the fix. I have applied it now.

PR #184, opened 2020-01-09
foomatic-rip: fix compilation with -fno-common

mwilck:
Starting from the upcoming GCC release 10, the default of -fcommon option will
change to -fno-common. This causes compilation errors in foomatic-rip. These
seem to be due to missing "external" declarations.

https://gcc.gnu.org/onlinedocs/gcc-9.2.0/gcc/Code-Gen-Options.html#index-fno-common


tillkamppeter:
Thanks for your patch. I have merged it now.

Issue #117, opened 2019-05-09
Graceful/Consistent handling of zero-page jobs


tillkamppeter:
It does not make much sense to send a file in a printable format but consisting of zero pages to a printer, but such jobs can easily appear, for example using the "page-ranges" option to select certain pages but the input file does not contain enough pages so that not even one page gets selected (like `page-ranges=3-4` on a one-page input file). Also errors in handling a PDF file manipulation utility can produce a PDF file not containing a single page.
If you print such a job the most consistent behavior would be that nothing gets printed on the printer (it is zero pages, not one empty page) and the job is taken from the queue without error.
To achieve this, all filters of cups-filters must behave appropriately, meaning that if the input data has zero pages that the output data must be a valid zero-page file in the output format of the filter and no input data at all needs to be considered as zero pages of input, too. Also a filter which is capable of removing pages from the job, like `pdftopdf`, needs to output a valid zero-page file if it removes all pages.
I have especially observed that `pdftopdf` generates illegal/broken PDF if one eliminates all the input pages via the `page-ranges` option. `gstoraster` and `mupdftoraster` produce zero bytes of output on a zero-page input file while `pdftoraster` produces a single blank page, which is all wrong. The correct output would be a zero-page Raster file which is simply only the 4-byte "magic string" of Raster, like "RaS2".
Also none of the filters accepts a totally empty file as input. It should be considered a zero-page file.
This problem was discovered when sending a PDF file with probably less than 3 pages and `page-ranges=3-4` to a print queue using HPLIP (`pdftopdf` -> `gstoraster` -> `hpcups` filter sequence) and the job caused an error (Issue #67).
Here the `hpcups` filter errored on totally empty (zero-byte) input. `pdftopdf` eliminated all pages and passed on broken PDF. Then `gstoraster` issued a warning and did not pass on any byte of output.
@smilingthax, @jberkenbilt, could you have a look into `pdftopdf` to make it always produce valid output, also in the zero-pages case? Could you also make it accepting totally empty input?
@tanmayanand44, could you take this also into account when working on the `pdftoraster` filter?

tanmayanand44:
@tillkamppeter Sure, I will do include this.

jberkenbilt:
[...]

PR #190, opened 2020-01-16
Added zero-page handling for texttopdf filter.

JaiLuthra1:


tillkamppeter:
Note that your check for the input being empty does not work, especially with stdin as input you cannot use lseek (how did you test). Also we want a totally empty file as zero-page PDF output. See issue #117. I have solved this now in another way, see commit 2db65a0c8.
But thanks anyway for your contribution.

PR #195, opened 2020-01-27
Add zero-page handling for pdftopdf

JaiLuthra1:


tillkamppeter:
Thank you very much for the patch, but on further filters make sure that it is distinguished between empty input and a broken, damaged, or wrong format input file, as in the former case the exit value has to be zero and in the latter case it has to be one. I fixed this now in commit 5c38121e.

PR #196, opened 2020-01-28
pdftops: added handling of empty input

JaiLuthra1:
pdftops filter exits with value 0 on empty input and zero page pdf and with value 1 on broken, damaged and wrong format input files and PS is considered a wrong format input for the pdftops filter to ensure similar behaviour for all filters accepting pdf as input.


tillkamppeter:
Thank you very much. I have merged it now and improved your pdf_pages() function.

PR #198, opened 2020-02-01
gstoraster : added handling of empty input

JaiLuthra1:


tillkamppeter:
Thank you for the patch, I have merged it now.
There were some small issues with it, once Ghostscript was called with a hard-coded "gs" instead of with CUPS_GHOSTSCRIPT and page counting on input coming from stdin (and buffered in a temporary file) did generally not work, as the buffer with the filename got overwritten. I have corrected this in commit 7503c2b26d3 now. This should also solve your problems with the wrapper scripts.

PR #199, opened 2020-02-03
added support for CJK fonts

dsam82:
Modified charset file and texttopdf filter to add support for CJK fonts.


tillkamppeter:
Thank you very much, I have merged your patch now.

PR #204, opened 2020-02-18
Fix crash in check_printer_with_option

zdohnal:
Initialize the value, add further checks, freeing memory and stop
allocating magic numbers.

Hi Till,

we got following report in Fedora https://bugzilla.redhat.com/show_bug.cgi?id=1802969 - cups-browsed was crashing in check_printer_with_option due uninitialized string.

I added several checks to the function code too.

Would you mind merging it?


tillkamppeter:
Thank you very much for the fix, I have merged it now.

Issue #206, opened 2020-02-22
foomatic-rip postscript preprocessing broken by #200


LynxChaus:
After fixing #200 - postscript preprocessing is broken. Removing return at filter/foomatic-rip/postscript.c:1081 fix it.

OdyX:
I can confirm this; foo2zjs doesn't print anything anymore since 1.27.1; that's a serious regression.

I have therefore included @LynxChaus 's patch and will send a pull-request when tested.

I'll change Debian's CUPS autopkgtests to detect these regressions too.

tillkamppeter:
What was the exact problem here? Did all PostScript files sent through foomatic-rip come out empty (zero bytes) instead of the document in the printer's format? Or did only zero-page PostScript files come out empty and the printer did not like an empty job in some form.

tillkamppeter:
And do all attempts to pass a PostScript file through foomatic-rip end up with a message "No pages left, outputting empty file." in the error_log?

OdyX:
The easiest way for me to reproduce this was to run this (with printer-driver-foo2zjs installed) with the latest test-drivers script from CUPS: https://salsa.debian.org/printing-team/cups/blob/debian/master/debian/tests/utils/test-drivers :
```
sudo ./debian/tests/utils/test-drivers -l "foo2zjs:0/ppd/foo2zjs/HP-LaserJet_Pro_P1102w.ppd"
```
Without this patch, I got:
```
* Driver foo2zjs:0/ppd/foo2zjs/HP-LaserJet_Pro_P1102w.ppd
 - Create test printer: done.
 - Print /usr/share/cups/data/topsecret.pdf: . Filter succeeded but resulting file is empty; aborting!
 - Interrupted (or finished), delete test printer: done.
```
. With the patch, I got:
```
* Driver foo2zjs:0/ppd/foo2zjs/HP-LaserJet_Pro_P1102w.ppd
 - Create test printer: done.
 - Print /usr/share/cups/data/topsecret.pdf: . done.
 - Print /usr/share/cups/data/form_russian.pdf: .. done.
 - Print /usr/share/cups/data/default.pdf: . done.
 - Print /usr/share/cups/data/secret.pdf: . done.
 - Print /usr/share/cups/data/standard.pdf: . done.
 - Print /usr/share/cups/data/unclassified.pdf: . done.
 - Print /usr/share/cups/data/classified.pdf: . done.
 - Print /usr/share/cups/data/form_english.pdf: .. done.
[...]

PR #207, opened 2020-02-22
Fix foomatic-rip postscript preprocessing

OdyX:
Fixes: #206 

Thanks to @LynxChaus for the patch proposal!


tillkamppeter:
Fixed correctly in Pull request #209.

PR #209, opened 2020-02-24
foomatic-rip: correct handling of jobs

JaiLuthra1:


tillkamppeter:
With this fix PostScript input is read completely and dropped into a temporary file before rendering/printing is started, simply to find out whether the file is not empty (zero pages). This breaks streaming of PostScript input. PostScript is a format which supports streaming and this we want to make use of for the Foomatic Printer Application to stream in PWG/Apple Raster input, by encapsulating it in PostScript and feeding this into `foomatic-rip`.

Reading the complete input to find out whether it contains pages is not needed, one simply feeds it into the Ghostscript command line with "bbox" output device line by line, until we have output (the "bbox" device produces two output lines for each page) and if we have output we have at least one page. So we only need to buffer the input up to here and then we simply kill Ghostscript and do the rendering (if we actually fond a page). After that the input can continue arbitrarily, so we can render even infinite jobs.

Here is the commit in the master branch: fd35106460

PR #214, opened 2020-03-02
Fix segfaults in test suite when test font is missing

zdohnal:
Hi Till,

Font team in Fedora changed naming of fonts in Fedora, so now default test font in test suite is not found, but test binaries or configure script do not check if it exists and binaries segfault instead.

Old font path:
/usr/share/fonts/dejavu/DejaVuSans.ttf

New font path:
/usr/share/fonts/dejavu-sans-fonts/DejaVuSans.ttf

I'm not sure if the change will be in other distros, so I fixed it by supplying new path to configure script, but the segfault if the font is not supplied does not seem as correct result.

The pull request fixes the segfault and adds the check in configure.ac - the configuration will end with error if the font file is not found.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -788,6 +788,9 @@ AC_ARG_WITH([test-font-path],
 	[with_test_font_path="$withval"],
 	[with_test_font_path="/usr/share/fonts/dejavu/DejaVuSans.ttf"]
 )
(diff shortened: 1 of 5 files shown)


zdohnal:
@tillkamppeter  Thank you, Till!

PR #218, opened 2020-03-12
cups-browsed: Added AllowResharingRemoteCUPSPrinters directive to cups-browsed

V1krant:
Hi Till,

The directive `NewBrowsePollQueuesShared` which shares printers discovered by the browse poll directive is added in the method `browse_poll_get_printers` as here it was already being checked for printers discovered by BrowsePoll directive.
Kindly let me know if this patch requires any modifications.

### DIFF
--- a/utils/cups-browsed.c
+++ b/utils/cups-browsed.c
@@ -430,6 +430,8 @@ static unsigned int BrowseInterval = 60;
 static unsigned int BrowseTimeout = 300;
 static uint16_t BrowsePort = 631;
 static browsepoll_t **BrowsePoll = NULL;
+static unsigned int NewBrowsePollQueuesShared = 0;
+static unsigned int AllowResharingRemoteCUPSPrinters = 0;
 static size_t NumBrowsePoll = 0;
 static guint update_netifs_sourceid = 0;
 static char local_server_str[1024];
@@ -7897,7 +7899,7 @@ gboolean update_cups_queues(gpointer unused) {
 	     * only if we have CUPS older than 2.2.
 	     * When you have remote queue, clean up and break from the loop.
 	     */
-	    if (p->netprinter != 0 || !HAVE_CUPS_2_2)
+	    if (p->netprinter != 0 || !HAVE_CUPS_2_2 || AllowResharingRemoteCUPSPrinters)
 	      ippDelete(cupsDoRequest(http, request, "/admin/"));
 	    else {
 	      ippDelete(request);
@@ -8465,7 +8467,8 @@ gboolean update_cups_queues(gpointer unused) {
 	    if (p->netprinter == 0 &&
 		strncmp(line, "*%", 2) &&
 		strncmp(line, "*PPD-Adobe:", 11) &&
-		ap_remote_queue_id_line_inserted == 0) {
+		ap_remote_queue_id_line_inserted == 0 &&
+              !AllowResharingRemoteCUPSPrinters) {
 	      ap_remote_queue_id_line_inserted = 1;
 	      cupsFilePrintf(out, "*APRemoteQueueID: \"\"\n");
 	    }
@@ -8605,6 +8608,12 @@ gboolean update_cups_queues(gpointer unused) {
 	num_options = cupsAddOption("printer-is-shared", "true",
 				    num_options, &options);
 	debug_printf("Setting printer-is-shared bit.\n");
+      } else if (NewBrowsePollQueuesShared &&
+      (val = cupsGetOption("printer-to-be-shared", p->num_options,
+               p->options)) != NULL) {
+	num_options = cupsAddOption("printer-is-shared", "true",
+				    num_options, &options);
+	debug_printf("Setting printer-is-shared bit.\n");
       } else {
 	num_options = cupsAddOption("printer-is-shared", "false",
 				    num_options, &options);
@@ -8617,7 +8626,7 @@ gboolean update_cups_queues(gpointer unused) {
        * network printer or if we have remote CUPS queue, do IPP request
        * only if we have CUPS older than 2.2.
        */
-      if (p->netprinter != 0 || !HAVE_CUPS_2_2)
+      if (p->netprinter != 0 || !HAVE_CUPS_2_2 || AllowResharingRemoteCUPSPrinters)
         ippDelete(cupsDoRequest(http, request, "/admin/"));
       else
         ippDelete(request);
@@ -10417,12 +10426,18 @@ found_cups_printer (const char *remote_host, const char *uri,
       (printer->domain == NULL || printer->domain[0] == '\0' ||
        printer->type == NULL || printer->type[0] == '\0')) {
     printer->is_legacy = 1;
+
     if (printer->status != STATUS_TO_BE_CREATED) {
       printer->timeout = time(NULL) + BrowseTimeout;
       debug_printf("starting BrowseTimeout timer for %s (%ds)\n",
 		   printer->queue_name, BrowseTimeout);
     }
   }
+
+  if (printer && NewBrowsePollQueuesShared &&
+      (HAVE_CUPS_1_6 || (!HAVE_CUPS_1_6 && !printer->is_legacy)))
+    printer->num_options = cupsAddOption("printer-to-be-shared", "true", printer->num_options, &(printer->options));
+
 }
 
 gboolean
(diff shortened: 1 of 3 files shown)


utils/cups-browsed.c
```diff
@@ -10712,6 +10713,12 @@ browse_poll_get_printers (browsepoll_t *context, http_t *conn)
       found_cups_printer (context->server, uri, location, info);
       printer = new_browsepoll_printer (uri, location, info);
       printers = g_list_insert (printers, printer, 0);
+      if (NewBrowsePollQueuesShared) {
+      int num_options = 0;
+	    num_options = cupsAddOption("printer-is-shared", "true",
+				    num_options, NULL);
+	    debug_printf("Setting printer-is-shared bit.\n");
+      }
     }
```
tillkamppeter: To which option set are you adding the "printer-is-shared" option here? "cupsAddOption()" needs the option set to add the option to as its 4th argument, where you set NULL. See the source code of cupsAddOption() in CUPS, file `cups/options.c`. Did you actually build and test this code?

V1krant: @tillkamppeter , I added this commit to get reviews from you about the code and therefore, did not completely test the code. However, I have added a new commit, after studying more about sending requests and changing options, which would change the `printer-is-shared` option when required.

tillkamppeter: This also is not the correct thing. Here you are setting the printer-is-shared bit on the remote print queue. Once this does not make sense as on the remote server all these printers are shared, otherwise you would not see them as a client. Also you are sending the request for setting the shared bit to your local CUPS, but the printers are on remote CUPS servers. The request will error.
Did you test what you are doing? With the new option set you should have your local queues for BrowsePolled printers shared, with the new option not set they should be not shared.

V1krant: @tillkamppeter , I am facing trouble while testing the code. When I try to modify the attribute
printer-is-shared of the queue on the client, it shows the error unauthorised. Even
when trying to change the printer-is-shared bit from web interface, the error is
`Cannot change printer-is-shared for remote queues`. I really want to solve this issue but I wanted your help to figure out why remote queue, instead of local queue is appearing on the client side. On the `cups-browsed.conf` file on the client, the following options are set.
BrowseRemoteProtocols dnssd cups
BrowseInterval 60
BrowseTimeout 300
BrowsePoll 192.168.43.210:631
OnlyUnsupportedByCUPS Yes

tillkamppeter: What is the device URI of the local queue with which you get the error (run `lpstat -v`)? Does it have a PPD file in `/etc/cups/ppd/`?

V1krant: The Device UR, which appears on the client is `implicitclass://D1_172_20_10_7/` where D1 is the description of the printer on the server. There also exists a ppd file for this printer.

V1krant: @tillkamppeter , I think my issue might be related to [this issue](https://github.com/apple/cups/issues/4738).

tillkamppeter: CUPS actually does not allow re-sharing print queues which point to a remote CUPS queue. CUPS recognizes these queues in two ways (search for `|= CUPS_PRINTER_REMOTE` in the `scheduler/printers.c` file of the CUPS source code):
- The device URI is CUPS-typical, like for example `ipp://host:631/printers/queue` (resource is `/printers/queue`, not `/ipp/print` or `/ipp/print/queue`) or `ipp://service_name._ipp._tcp.local/cups` (resource `/cups` instead of no resource).
- The PPD file contains `*APRemoteQueueID: ...`.

Long time ago I followed the CUPS functionality by letting cups-browsed add a `*APRemoteQueueID: ...` line to the PPD. Simply search for "`*APRemoteQueueID:`" in `utils/cups-browsed.c` and comment out these lines for your further studies. Then restart cups-browsed. Now you should be able to share such a queue.
Note that all queues discovered by BrowsePoll are remote CUPS queues, so the cups-browsed as it is now will always add `*APRemoteQueueID: ...` to the PPD file.
There are two solutions: Either remove the code for adding `*APRemoteQueueID: ...` altogether or add a second boolean directive to `cups-browsed.conf`, call it "AllowResharingRemoteCUPSPrinters" and if set to "Yes" let the `*APRemoteQueueID: ...` not be added. Your directive will only take effect then.

V1krant: > add a second boolean directive to `cups-browsed.conf`, call it "AllowResharingRemoteCUPSPrinters" and if set to "Yes" let the `*APRemoteQueueID: ...` not be added.

@tillkamppeter , In that case, I've observed that print queues are automatically shared if `AllowResharingRemoteCUPSPrinters` is set to yes, may be because the option settings are copied of the shared printer. Therefore, there wouldn't be a need of the `NewBrowsePollQueuesShared` directive. But should I add it in case someone does not want to share the queues browsepolled from a server?

V1krant:
> add a second boolean directive to `cups-browsed.conf`, call it "AllowResharingRemoteCUPSPrinters" and if set to "Yes" let the `*APRemoteQueueID: ...` not be added.

@tillkamppeter , In that case, I've observed that print queues are automatically shared if `AllowResharingRemoteCUPSPrinters` is set to yes, may be because the option settings are copied of the shared printer. Therefore, there wouldn't be a need of the `NewBrowsePollQueuesShared` directive. But, do you think it should be added in case someone does not want to share the queues browsepolled from a server?
As for now, I've tested and added the code for `AllowResharingRemoteCUPSPrinters` directive.

V1krant:
@tillkamppeter , Should I squash the commits for this to be merged?

tillkamppeter:
Note that all discovered remote CUPS queues are shared on their CUPS servers., otherwise they would not be discoverable. I do not know why now all client queues get shared when we allow re-sharing by the cups-browsed code. Perhaps this is a local default on your client.
The `AllowResharingRemoteCUPSPrinters` directive should only influence whether `*APRemoteQueueID: ...` is added to the PPD or not, and so whether the user can switch the printer-is-shared bit to the queue being shared.
We still need the `NewBrowsePollQueuesShared` directive, so that one has active control on whether a local queue newly created by cups-browsed triggered by BrowsePoll is shared or not. Default should be not shared. Please always actively set a new queue to shared or not shared, so that it follows this directive. Add to the documentation that the `AllowResharingRemoteCUPSPrinters` must be set to `Yes` so that new BrowsePoll queues can get shared.

V1krant:
> I do not know why now all client queues get shared when we allow re-sharing by the cups-browsed code. Perhaps this is a local default on your client.

Apparently, they were shared because in `update_cups_queues`, the request is not sent to server as `printer->netprinter` is verified not to be 0. But if `AllowResharingRemoteCUPSPrinters` is set, the request is sent and is accepted without an error.

V1krant:
@tillkamppeter , I've tested and added the `NewBrowsePollQueuesShared` directive. I've used the option `printer-to-be-shared` and not `printer-is-shared` so that it doesn't create problems regarding creation/modification of local queues in `update_cups_queues`. Please let me know if any further changes are required.

tillkamppeter:
Have you taken care that the `NewBrowsePollQueuesShared` directive should only influence local print queues based on remote CUPS printer discovery via BrowsePoll and not on local queues based on legacy CUPS broadcasting of remote CUPS queues by old-generation (1.5 and older) CUPS servers?

V1krant:
@tillkamppeter , I have taken care of that in my latest commit.

utils/cups-browsed.c
```diff
@@ -430,6 +430,8 @@ static unsigned int BrowseInterval = 60;
 static unsigned int BrowseTimeout = 300;
 static uint16_t BrowsePort = 631;
 static browsepoll_t **BrowsePoll = NULL;
+int NewBrowsePollQueuesShared = 0;
```
tillkamppeter: @V1krant, could you add the "static" here? Thanks.

V1krant: Yeah sure.

tillkamppeter:
@V1krant, OK, we are shortly before merging your Pull Request. Could you adjust the indentation of your code with the already existing code, the tabs in the file should stand for 8 spaces, or better, use only spaces.
Could you then also squash your commits to one?
Did you also test your code for all situations?

V1krant:
> @V1krant, OK, we are shortly before merging your Pull Request. Could you adjust the indentation of your code with the already existing code, the tabs in the file should stand for 8 spaces, or better, use only spaces.
> Could you then also squash your commits to one?

Sure Till, I'll just update it and push the commit.

> Did you also test your code for all situations?

I've tested it for all situations except for old versions old cups (for the `HAVE_CUPS_1_6` flag).

V1krant:
@tillkamppeter , I've squashed the commits and replaced tabs with spaces. Kindly let me know if any further changes are required

utils/cups-browsed.c
```diff
@@ -8462,13 +8464,15 @@ gboolean update_cups_queues(gpointer unused) {
...
-	    }
+    if (!AllowResharingRemoteCUPSPrinters) {
+        if (p->netprinter == 0 &&
+            strncmp(line, "*%", 2) &&
+            strncmp(line, "*PPD-Adobe:", 11) &&
+            ap_remote_queue_id_line_inserted == 0) {
+          ap_remote_queue_id_line_inserted = 1;
+          cupsFilePrintf(out, "*APRemoteQueueID: \"\"\n");
+        }
+      }
```
tillkamppeter: Could you adjust the indentation of this part, so that it matches the surrounding code? All the rest seems OK.

V1krant: @tillkamppeter, Does it look good now?

tillkamppeter: As the existing code uses tabs which represent 8 spaces you need to have your editor display the code with the tabs representing 8 spaces and then adjust your code to the existing one. For me it looks like that your editor is showing tabs as only 2 or 4 spaces.

V1krant: > For me it looks like that your editor is showing tabs as only 2 or 4 spaces.

You are right. I've changed my editor's configuration now. Could you review it now please?

tillkamppeter: Now it is all OK, thanks.

V1krant: Thanks @tillkamppeter.

tillkamppeter:
@V1krant, thank you very much. I merged your work now.

ibaldonl:
Hey! Thanks so much for your work guys!!! Thanks for contributing it. If you ever come to Uruguay let me know!

PR #219, opened 2020-03-13
Fix possible memory leaks in cups-browsed daemon

zdohnal:
Hi Till,

I found several other possible memory leaks in cupsAddOption() calls - using strdup as parameter of cupsAddOption() results in losing the allocated pointer, because cupsAddOption() allocates the memory by itself and since the pointer allocated by strdup is not saved anywhere, it leaks.

Plus I assigned NULL into response pointer too, because it is in the loop and the garbage after freeing can confuse the conditional when getting new response fails.

I tested it in Fedora rawhide and I was able to see remote print queues and print to one of them.

Would you mind adding the fix into the project?

Thank you in advance,

Zdenek

### DIFF
--- a/utils/cups-browsed.c
+++ b/utils/cups-browsed.c
@@ -6278,6 +6278,7 @@ on_job_state (CupsNotifier *object,
 	    }
 
 	    ippDelete(response);
+            response = NULL;
 	  } else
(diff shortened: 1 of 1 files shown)


tillkamppeter:
Merged. Thank you very much for your contribution.

PR #220, opened 2020-03-15
driverless: Unlink temporary file

hansmi:
The call to unlink(2) in "driverless" used the wrong variable,
triggering a silent failure (ENAMETOOLONG), and leaving a temporary file
behind.


tillkamppeter:
Thank you very much for your patch, I have merged it now.

PR #222, opened 2020-04-02
Replace '==' in configure.ac test with '='

dcantrell:
== is a bash-specific test operator.  Use = here for portability.


tillkamppeter:
Thank you very much for the fix. I have merged it now.

PR #226, opened 2020-04-08
Fix memory issues in ppdgenerator, cups-browsed

zdohnal:
Hi Till,

I did an other round of valgrind testing of cups-browsed and found out several other issues with memory (leaks, conditional jump on uninitialized value), which the pull request fixes.

Would you mind adding it into the project?


tillkamppeter:
Thank you very much. I have merged your fixes now.

zdohnal:
@tillkamppeter Thank you!

PR #229, opened 2020-05-06
Update ppdgenerator.c

mohitmo:
During installation of cups-filter, a warning showed due to freeing const char*. This will fix it.


tillkamppeter:
Thank you very much for the fix, I have merged it now.

PR #230, opened 2020-05-07
Do not pass NULL to fprintf()

kucharskim:
Passing NULL to printf-family of functions is undefined
behaviour. OpenBSD has detection mechanism for occurrences
of such condition. On my machine I get following messages
in syslog, while printing:

vfprintf %s NULL in "DEBUG: OUTFORMAT="%s", so output format will be %s "

Changes in this diff make those messages go away.


tillkamppeter:
Thank you very much, I have merged your patch now.

PR #236, opened 2020-05-21
Map two-sided-short-edge to DuplexTumble

IslandCow:
two-sided-short-edge and two-sided-long-edge were both mapped to
DuplexNoTumble.  DuplexNoTumble should be mapped to the long-edge
tumble and short-edge to DuplexTumble.


tillkamppeter:
Thanks for the patch, I have merged it now.

PR #240, opened 2020-05-27
configure.ac: Use AS_IF instead of AC_CHECK_FILE for font check

### LINKED ISSUE #239: 1.27.4 fail to cross-compile (configration phase)
```
configure:22805: result: yes
configure:23532: checking for "/usr/share/fonts/truetype/dejavu/DejaVuSans.ttf"
configure:23538: error: cannot check for file existence when cross compiling
```

It is due to the `AC_CHECK_FILE` macro: https://github.com/OpenPrinting/cups-filters/blob/master/configure.ac#L791

zdohnal:
Fixes #239 - AS_IF has support for cross-platform compilation.


tillkamppeter:
Thank you very much, I have merged your fix now.

PR #244, opened 2020-06-16
cups-browsed.8: Remove mentions of README and AUTHORS files

zdohnal:
Hi,

man page of cups-browsed mentions the absolute path to README and AUTHORS files, which can be different due packaging the project.
Currently, I have a downstream patch in Fedora for renaming the path and Ubuntu seems to duplicate README, NEWS, copyright files to both packages - cups-browsed and cups-filters - just into different dirs, which seems unnecessary for doc files.

In my opinion it would be the best just to remove the misleading entries from cups-browsed man page and stop duplicating files/stop carrying a downstream patch. The doc files will be available once in main location, /usr/share/doc/cups-filters.

Would it be okay?


tillkamppeter:
Thanks for the fix. I have merged it now.

PR #245, opened 2020-06-17
modified when to call cupsGetNamedDest()

mohitmo:
This PR modifies when to call cupsGetNamedDest(). 
Now instead of calling cupsGetNamedDest() for every printer , we first do a get-printer-attribute IPP request on local cupsd and check if we have CUPS queue with that name. If found then we call cupsGetNamedDest().

### DIFF
--- a/utils/cups-browsed.c
+++ b/utils/cups-browsed.c
@@ -7833,10 +7833,10 @@ gboolean update_cups_queues(gpointer unused) {
         else {
           debug_printf ("CUPS queue with URI %s found, but with \"printer-name\" %s", uri, ippGetString(attr, 0, NULL));
         }
+
+        ippDelete(response);
       }
 
-      ippDelete(response);
-   
       if (dest) {
 	/* CUPS has found a queue with this name.
 	   Either CUPS generates a temporary queue here or we have already
--- a/utils/cups-browsed.c
+++ b/utils/cups-browsed.c
@@ -7807,7 +7807,35 @@ gboolean update_cups_queues(gpointer unused) {
 #ifdef HAVE_CUPS_1_6
       /* Check whether there is a temporary CUPS queue which we would
          overwrite */
-      dest = cupsGetNamedDest(http, p->queue_name, NULL);
+      ipp_t *response =  NULL;
+      const char const *pattrs[] =
+                {
+                  "printer-name"
+                };
+
+      const char const *req_attrs[] =
+                {
+                  "printer-name"
+                };
+
+      if ((response = 
+           get_printer_attributes2(http, uri, 
+                       pattrs, 1, req_attrs, 1, 0)) == NULL) {
+        debug_printf ("No CUPS queue with URI %s found\n", uri);
+      }
+      else {
+        attr = ippFindAttribute(response, "printer-name", IPP_TAG_NAME);
+        if(!strcmp(ippGetString(attr, 0, NULL), p->queue_name)){
+          dest = cupsGetNamedDest(http, p->queue_name, NULL);
+          debug_printf ("CUPS queue %s with URI %s found", p->queue_name, uri);
+        }
+        else {
+          debug_printf ("CUPS queue with URI %s found, but with \"printer-name\" %s", uri, ippGetString(attr, 0, NULL));
+        }
+      }
+
+      ippDelete(response);
+   
       if (dest) {
 	/* CUPS has found a queue with this name.
 	   Either CUPS generates a temporary queue here or we have already
--- a/utils/cups-browsed.c
+++ b/utils/cups-browsed.c
@@ -7807,6 +7807,7 @@ gboolean update_cups_queues(gpointer unused) {
 #ifdef HAVE_CUPS_1_6
       /* Check whether there is a temporary CUPS queue which we would
          overwrite */
+      dest = NULL;
       ipp_t *response =  NULL;
       const char const *pattrs[] =
                 {


utils/cups-browsed.c
```diff
@@ -7807,7 +7807,35 @@ gboolean update_cups_queues(gpointer unused) {
 #ifdef HAVE_CUPS_1_6
       /* Check whether there is a temporary CUPS queue which we would
          overwrite */
-      dest = cupsGetNamedDest(http, p->queue_name, NULL);
+      ipp_t *response =  NULL;
```
tillkamppeter, line 7810: You should set "dest = NULL" before this line, so that if the get-printer-attributes request is without result or with bad result that dest is not on a random value or on the value of the previous CUPS queue updated, so that nothing wrong can happen on the "if (dest)" below.

mohitmo: Thank You!
I have assigned dest = NULL.

zdohnal:
Hi @mohitmo and @tillkamppeter ,

I think the patch would rule out all cups temporary queues, which will become permanent with cups-browsed, but I'll test it. Is it intentional?

Either way, 

```
ippDelete(response)
```

should be called only when response is not NULL - you can do it by moving it into '''else''' block where you know response isn't NULL or just check it with '''if'''

utils/cups-browsed.c
```diff
@@ -7807,7 +7807,35 @@ gboolean update_cups_queues(gpointer unused) {
...
+        if(!strcmp(ippGetString(attr, 0, NULL), p->queue_name)){
+          dest = cupsGetNamedDest(http, p->queue_name, NULL);
+          debug_printf ("CUPS queue %s with URI %s found", p->queue_name, uri);
+        }
+        else {
+          debug_printf ("CUPS queue with URI %s found, but with \"printer-name\" %s", uri, ippGetString(attr, 0, NULL));
+        }
+      }
+
+      ippDelete(response);
```
zdohnal, line 7838: Here what I meant in the comment.

tillkamppeter:
@zdohnal, it depends on the configuration in `cups-browsed.conf`, whether we make temporary CUPS queues permanent or leave them alone. Making them permanent is for supporting legacy print dialogs which by themselves do not support temporary CUPS queues.
In `cups-browsed.conf` you set
```
OnlyUnsupportedByCUPS Yes
```
if you do not want that cups-browsed creates permanent queues where CUPS would already create temporary ones. The setting influences the get_local_printers() function. The list created by the function determines if a discovered printer is rejected right away as there is already a local queue for it. So we reach the code patched here only if we allow overwriting temporary CUPS queues.

zdohnal:
@tillkamppeter .
OnlyUnsupportedByCUPS has the default value - 'No'. 
With the patch, whole '''if (dest)''' block is skipped, because there is no such uri (e.g. ipp://localhost/printers/HP_LaserJet_M1536dnf_MFP_42307C_) on localhost before someone creates it - which is the case of temporary queues.
The patch results into replacing the temp queue instead of taking it over via setting/resetting printer-is-shared attribute, ignoring whether it has any jobs in the queue or different ppd file.

Is it expected? I would say a different ppd wouldn't matter because of the current state of CUPS project, but users wouldn't be happy if there were any jobs on the queue and cups-browsed just replaced it.

tillkamppeter:
@mohitmo, did you test whether the get-printer-attributes IPP request really replaces correctly the cupsGetNamedDest() call, meaning that in the case that CUPS would create a temporary queue but did not actually create one that we get a positive answer of CUPS? Can you test that?

mohitmo:
> @mohitmo, did you test whether the get-printer-attributes IPP request really replaces correctly the cupsGetNamedDest() call, meaning that in the case that CUPS would create a temporary queue but did not actually create one that we get a positive answer of CUPS? Can you test that?

I created a temporary queue with lpstat and lp, and after that ran cups-browsed, and it found the temporary queue with its URI and called cupsGetNamedDest() correctly.
Do I have to try something else?

zdohnal:
@mohitmo This way you create temporary queue on localhost for time being, but it is not the default state.
Try to test it only when you have a printer turned on, without any lp or lpstat command. Even such queue is reachable via cupsGetNamedDest().

mohitmo:
> @mohitmo This way you create temporary queue on localhost for time being, but it is not the default state.
> 
> To put it simply cups temp queue is a permanent queue which will be removed after some time/action. Your testing just fell into 'some time' frame, when the temp queue was available on localhost.
>
 
In this way it does not find a queue with the URI, and creates permanent queue.

> Try to test it only when you have a printer turned on, without any lp or lpstat command. Even such queue is reachable via cupsGetNamedDest().
It is because cupsGetNamedDest() will create temporary queue, if it doesn't already exist.

zdohnal:
> > To put it simply cups temp queue is a permanent queue which will be removed after some time/action. Your testing just fell into 'some time' frame, when the temp queue was available on localhost.
> 
> In this way it does not find a queue with the URI, and creates permanent queue.
> 

Not finding a queue with URI on localhost:631 is expected for temp queue because it isn't permanent queue, but the temp queue still exists within destinations, but it shouldn't create permanent queue as a new queue, but reset/set printer-is-shared attribute to take over the temp queue.

> > Try to test it only when you have a printer turned on, without any lp or lpstat command. Even such queue is reachable via cupsGetNamedDest().
> 
> It is because cupsGetNamedDest() will create temporary queue, if it doesn't already exist.

Based on Avahi, looks good to me.

tillkamppeter:
AFAIK cupsGetNamedDest() does not create a temporary queue, only cupsCopyDestInfo() does.

What we want to know at the point where the cupsGetNamedDest() was is whether CUPS WOULD create a temporary queue, not whether CUPS HAS created a temporary queue.

If for a given discovered printer CUPS would create a temporary queue, we want
- in case of `OnlyUnsupportedByCUPS Yes` skip creating a local permanent queue altogether. This is done at another place in the code of cups-browsed.
- in case of `OnlyUnsupportedByCUPS No` (the default) we want to create a local permanent queue which is equivalent to the temporary queue which CUPS would create, but it is permanent, not disappearing after a minute without use. Therefore we trigger CUPS to actually create this queue (cupsCopyDestInfo() call), download the queue's PPD file, and after that try to make the queue permanent in minimum-invasive way by setting its printer-is-shared bit and releasing the printer-is-shared bit again, and if this fails we remove the queue and re-create it with the save PPD file.

So if the get-printer-attributes IPP request does not detect the case whether CUPS would create a temporary queue, but cupsGetNamedDest() does, we need to investigate the source code of the cupsGetNamedDest() function whether we can extract a part of the function which is much quicker and gives the desired result.

Another possible improvement: We should perhaps only download the PPD of the temporary CUPS queue if we have actually failed to make it permanent via the printer-is-shared bit.

zdohnal:
> AFAIK cupsGetNamedDest() does not create a temporary queue, only cupsCopyDestInfo() does.
> 

Sorry, actually didn't check @mohitmo statement, I thought it was true.

> What we want to know at the point where the cupsGetNamedDest() was is whether CUPS WOULD create a temporary queue, not whether CUPS HAS created a temporary queue.
> 
> If for a given discovered printer CUPS would create a temporary queue, we want
> 
>     * in case of `OnlyUnsupportedByCUPS No` (the default) we want to create a local permanent queue which is equivalent to the temporary queue which CUPS would create, but it is permanent, not disappearing after a minute without use. Therefore we trigger CUPS to actually create this queue (cupsCopyDestInfo() call), download the queue's PPD file, and after that try to make the queue permanent in minimum-invasive way by setting its printer-is-shared bit and releasing the printer-is-shared bit again, and if this fails we remove the queue and re-create it with the save PPD file.
> 

Which doesn't happen after the patch.

> 
> So if the get-printer-attributes IPP request does not detect the case whether CUPS would create a temporary queue, but cupsGetNamedDest() does, we need to investigate the source code of the cupsGetNamedDest() function whether we can extract a part of the function which is much quicker and gives the desired result.
>

It is because you use non-existing uri for get-printer-attributes request.
* before get-printer-attributes request, you assemble uri based on protocol (ipp), hostname (localhost), port (631) and resource ('/printers/' and queue name from avahi)
* then you do get-printer-attributes request, which fails
* fails because such uris have only currently installed queues

It worked for @mohitmo , because his temp queue wasn't cleaned up after lp/lpstat command, so it was available on the localhost for some time.

If you would want that IPP request have the same functionality as ```cupsGetNamedDest()```, you would need to communicate with Avahi as ```cups_enum_dests()``` does.

I'm sorry @mohitmo  @tillkamppeter , would you mind explaining why the new IPP request is needed? I'm concerned about performance - it is one more IPP request, which can slow down cups-browsed. Wouldn't it be possible to stay with ```cupsGetNamedDest()``` and then check IPP attribute for uri? (Break when it is on localhost, continue if it isn't)

mohitmo:
> I'm sorry @mohitmo @tillkamppeter , would you mind explaining why the new IPP request is needed? I'm concerned about performance - it is one more IPP request, which can slow down cups-browsed. Wouldn't it be possible to stay with `cupsGetNamedDest()` and then check IPP attribute for uri? (Break when it is on localhost, continue if it isn't)

Actually, as per the testing done by me to find how much time is consumed by a function, I found that update_cups_queues() took a large amount of time. Upon further investigation I found that cupsGetNamedDest() takes a large amount of time.

zdohnal:
> @zdohnal Actually, as per the testing done by me to find how much time is consumed by functions of cups-browsed, I found that update_cups_queues() took a large amount of time. Upon further investigation I found that cupsGetNamedDest() takes a large amount of time.

Good to know, thank you! Does cupsGetNamedDest() takes a lot of time by default or only with specific queues?
My guess would be that requests on avahi will take most of time - so maybe call cupsGetNamedDest() only for temp queues would make sense?

mohitmo:
> Good to know, thank you! Does cupsGetNamedDest() takes a lot of time by default or only with specific queues?
@zdohnal It takes a lot of time by default. I also tried to replace it with cupsGetDests2() and cupsEnumDests(), but both were taking more time than cupsGetNamedDest(). 
> My guess would be that requests on avahi will take most of time - so maybe call cupsGetNamedDest() only for temp queues would make sense?
How do you suggest to do that?

zdohnal:
> > My guess would be that requests on avahi will take most of time - so maybe call cupsGetNamedDest() only for temp queues would make sense?
> 
> How do you suggest to do that?

Temp queues are not found on localhost by default, so if your IPP request fails, do cupsGetNamedDest(). Then go into ```if (dest)``` block only when cupsGetNamedDest() returns not-NULL.

tillkamppeter:
@mohitmo, I have another idea where one could try:

Have a look at the function get_local_printers(). It generates a list of the local CUPS' print queues, including potential temporary queues depending on whether OnlyUnsupportedByCUPS is set or not.

The resulting list is used to check whether our attached local CUPS already has a queue with the name of our new queue and cancel queue creation if so. This is checked in an early stage after discovery of a remote printer via Avahi.

For the code piece in update_cups_queues() which we are discussing serves for checking whether our new queue would replace a temporary CUPS queue.

In the case of OnlyUnsupportedByCUPS=Yes we already have rejected the discovered remote printer in question earlier, so we do treat it at all in update_cups_queues(). So in this case running this code piece does not make sense. So we can accelerate by putting it into
```
if (OnlyUnsupportedByCUPS == 0) {

}
```
In case of OnlyUnsupportedByCUPS=No we should already let get_local_printers() register the (potential) temporary queues of the local CUPS in a separate list (is currently only done with OnlyUnsupportedByCUPS=Yes), so we do not need to call cupsGetNamedDest() to determine whether CUPS would create a remote queue.

So what we should do is:
1. Let get_local_printers() always behave as it now behaves with OnlyUnsupportedByCUPS=Yes, always registering printers for which CUPS would create a temporary queue in cups_supported_remote_printers, no distinction by OnlyUnsupportedByCUPS.
2. Put the code piece in update_cups_queues() which we are discussing here into
```
if (OnlyUnsupportedByCUPS == 0) {

}
```
3. Replace the get-printer-attributes IPP request which did not work for us by checking whether our printer is listed in cups_supported_remote_printers and only if yes, proceed with triggering the creation of the temporary queue, downloading its PPD, setting/unsetting its printer-is-shared bit, ...

This way we should reduce the cupsGetNamedDest() calls to a minimum. Please try out this.

mohitmo:
@tillkamppeter I did what you had asked. But it was taking similar time as with calling cupsGetNamedDest() with every printer.
 I then also tried pre-loading cupsEnumDests() before main for loop, and using cupsGetDest() instead of cupsGetNamedDest(), and it takes less time than cupsGetNamedDest(). 
Shall I create a PR?

PR #247, opened 2020-06-19
Correction default color

Ordissimo:
Correction of the default value of the color, after configuration of the CANON in IPP the test page is printed in black and white. The test empties the cartridges!
A more logical choice is Color.
Reproduced with CANON TS3150, TS3350, TS5150


Ordissimo:
Thank's

Issue #241, opened 2020-06-08
'KeepGeneratedQueuesOnShutdown' concerns     


zdohnal:
Hi,

I saw the latest cups-filters version has 'KeepGeneratedQueuesOnShutdown' functionality and I would like to express my concerns about implementing the feature and possible problems.

1) turning off cups-browsed will not clean the queues which are created by cups-browsed - IMO cleaning print queues is the same behavior as cupsd has - when you stop cupsd, there are no queues available, so cups-browsed should behave the same
2) the feature would make sense if cups-browsed was doing only device discovery and print queue creation, but IIUC it handles dbus notifications too - if cups-browsed is not started because of whatever reason, implicitclass backend will not work if the default destination uri will not work or when there is no default destination uri at all (e.g. cups-browsed ended before it was able to set destination).

What do you think about it? IMO the feature can bring up more serious issues than it solves (duplicate notifications don't seem as a serious issue comparing to non-functional queue).

zdohnal:
And 'BrowsePoll' directive is affected too - if you run cups-browsed with BrowsePoll set, cups-browsed will populate the print queue list. Then remove BrowsePoll and restart cups-browsed -> print queues from BrowsePoll should disappear, but they are still around.

tillkamppeter:
First, the new behavior of keeping the queues on shutdown is optionsl, one can get back to the old behavior by setting `KeepGeneratedQueuesOnShutdown No` in `cups-browsed.conf`.

The queues stay conserved as long as cups-browsed is not running. When restarting cups-browsed the queues are recognized by cups-browsed as coming from cups-browsed and considered unconfirmed. If the corresponding printer is not re-discovered within a timeout, the queue is removed.

The idea is that short interruptions of cups-browsed, for example for a log rotate, will not require a re-creation of the queues which produces annoying notification pop-ups. Also if you shut down your machine in the evening and restart it the next morning queues will not get re-created with notifications.

If queues of printers which do not get re-discovered after the restart of cups-browsed do not get removed after a timeout, this is an independent bug which was already there before (but before only got visible after a crash or a hard kill (`kill -9 ...`) of cups-browsed).

The queues are marked as coming from cups-browsed. If you shut down cups-browsed and after that cups, you can see this in `/etc/cups/printers.conf`. The option is `cups-browsed`.

[...]

PR #252, opened 2020-06-23
configure.ac: Add configure option for local queues naming for remote…

zdohnal:
… CUPS queues

Hi Till,

there's the first pull request for https://github.com/OpenPrinting/cups-filters/issues/241#issuecomment-641413728 .


zdohnal:
Included in #253 .

PR #253, opened 2020-06-23
[WIP] configure.ac: Add configure option for keeping created queues during shutdown

zdohnal:
Hi Till,

there is another pull request mentioned in https://github.com/OpenPrinting/cups-filters/issues/241#issuecomment-641413728 .

It introduces an option --enable-saving-created-queues, which sets the default value for KeepGeneratedQueuesOnShutdown in cups-browsed.conf. The default is 'no'.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -840,6 +840,30 @@ PKG_CHECK_EXISTS([liblouis], [
 AM_CONDITIONAL(ENABLE_BRAILLE, test "x$enable_braille" = xyes)
 AC_SUBST(TABLESDIR)
 
+# =========================================
+# Local queue naming for remote CUPS queues
+# =========================================
+AC_ARG_WITH([remote-cups-local-queue-naming],
+	[AS_HELP_STRING([--with-remote-cups-local-queue-naming=DNS-SD|MakeModel|RemoteName], [Choose the origin of local queue naming for remote CUPS queues, default based on DNS-SD ID])],
+	[case "x$withval" in
+		"xDNS-SD")
+			REMOTE_CUPS_LOCAL_QUEUE_NAMING="DNS-SD"
+			;;
+		"xMakeModel")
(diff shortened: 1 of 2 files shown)


zdohnal:
Damn, I wanted to have two different PR for two features, but avoid merge issues... I'll close the previous PR, since this PR has its commit too.

You can still review two separate commits though.

zdohnal:
The first commit adds ```--with-remote-cups-local-queue-naming```, which can have values - ```DNS-SD```, ```MakeModel``` or ```RemoteName```. The default is ```DNS-SD```, which is set when our configure option is not set or when it is set with invalid value.

I did a little indenting fix during options printing during configure too.

tillkamppeter:
Thank you for the pull request, but to be consistent with the already existing cups-browsed code build-time default changes should not be executed by modifying cups-browsed.conf during build but by modifying the initialization of the appropriate global variables (LocalQueueNamingRemoteCUPS, KeepGeneratedQueuesOnShutdown) in utils/cups-browsed.c. See there how I did it for other variables.

zdohnal:
Thank you for the review!

I did the pull request based on how you handle ```BrowseRemoteProtocols``` directive - is the handling incorrect?

If you have a preference how to handle a specific use case, please provide a contributor guidelines at least, if you don't want to change the old code.

tillkamppeter:
`BrowseRemoteProtocols` is the only exception (perhaps I should change it to make it consistent with the rest), take for example `CreateIPPPrinterQueues` as an example of how to do it.

zdohnal:
@tillkamppeter cups-filters is not compilable after libppd moving in - compilation is trying to fetch headers from CUPS, not from cups-filters.

tillkamppeter:
I have now done improvements on libppd to better separate it from libcups, renaming all private functions, types, internal macros. Please try again.

zdohnal:
I put the changes into other branch, so I created a new pull request. Closing this one.

PR #256, opened 2020-06-26
Potential null dereference fix (#1)

DuratarskeyK:
Fix a potential NULL dereference, when PPD file isn't present or can't be openned


tillkamppeter:
Thank you very much. I have merged your patch now.

PR #258, opened 2020-07-03
fix: get inheritable attribute from parent

liuli0217:
when /mediabox is inherited from parent, can't get rect. the function getInheritableFieldValue(key) from QPDF dosen't know if key is inheritable.


tillkamppeter:
Thank you very much for your fix, I have merged it now.

PR #264, opened 2020-07-15
pclmtoraster: Completed implementation of remaining features

V1krant:
The following features have been added.
1. Support for banded and planar rasters and support for conversion of bits.
2. Support for Duplex printing
3. Margins and page sizes would be matched from ppd (just as in pdftoraster with slight modifications).
4. Support for bi-level output rasters.

Test reports are available [here](https://github.com/V1krant/test-report). (Size ~500 mb).


tillkamppeter:
Is it not possible to make revtable and dither1 part of the public API of libcupsfilters so that these two can also get removed from pdftoraster?

tillkamppeter:
Or alternatively, you move all functions of pdftoraster which use revtable and dither1 into libcupsfilters, even if they are currently only used by pdftoraster and not by pclmtoraster.

tillkamppeter:
You copy over the margin modifications from pdftoraster, you could perhaps put them into libppd so that both filters (and further filters or Printer Applications in the future) could use them. Add the new function(s) also to ppd.h then as they will be part of libppd's API.

tillkamppeter:
Otherwise your changes look nice to me. Thanks for your great work.

V1krant:
> Is it not possible to make revtable and dither1 part of the public API of libcupsfilters so that these two can also get removed from pdftoraster?

Actually, both of these are used by some functions which are only in pdftoraster and not in pclmtoraster, therefore I did not copy them.
> Or alternatively, you move all functions of pdftoraster which use revtable and dither1 into libcupsfilters, even if they are currently only used by pdftoraster and not by pclmtoraster.

Ok, I'll move the corresponding functions to libcupsfilters.

> You copy over the margin modifications from pdftoraster, you could perhaps put them into libppd so that both filters (and further filters or Printer Applications in the future) could use them. Add the new function(s) also to ppd.h then as they will be part of libppd's API.

Thanks for the suggestion Till. I'll move them to libppd and will update the code.

> Otherwise your changes look nice to me. Thanks for your great work.

Thanks Till. Thanks to you and @sahilarora535 too for guiding me and helping me with the project.

tillkamppeter:
What looks strange for me is that your `ppdRasterMatchPPDSize()` returns the matching sizes as `ppd_size_t*` but where you call this function you do not drop this return value into a variable. How do you obtain the result?

V1krant:
> What looks strange for me is that your `ppdRasterMatchPPDSize()` returns the matching sizes as `ppd_size_t*` but where you call this function you do not drop this return value into a variable. How do you obtain the result?

Actually, I thought of returning the matched size and then apply the changes to the header, earlier. But later on, this was not needed and the values were stores in the dimensions and margins array. However, I forgot to change the return value from `ppd_size_t*` to `void`. I'll do it now.

V1krant:
I've also added the landscape variable, that can be used to know if the matched size is in landscape orientation.

ppd/raster-interpret.c
```diff
@@ -514,6 +514,17 @@ ppdRasterMatchPPDSize(
...
   char  	pageSizeRequested[64];  /* Requested PageSize */
 
+  if (!header)
+  {
+    _ppdRasterAddError("Page header cannot be NULL!\n");
+    return (-1);
+  }
+  if (!ppd)
+  {
+    _ppdRasterAddError("PPD file not found!\n");
```
tillkamppeter, line 524: Note that here we do not have a case of a file not being found but PPD info not being supplied, better use "PPD info not supplied!\n" here.

tillkamppeter:
Looks now all OK for me. Could you fix the error message as I mentioned earlier?

V1krant:
> Note that here we do not have a case of a file not being found but PPD info not being supplied, better use "PPD info not supplied!\n" here.
> > Looks now all OK for me. Could you fix the error message as I mentioned earlier?

Till, I've changed the error message to "PPD info not supplied!\n". Were you referring to something else?

Also, should I squash the commits into one, or should I leave them as it is for reference?

tillkamppeter:
I have merged it now, thank you very much.

Could you have a look at the following warning and fix it:
```
cupsfilters/bitmap.c: In function ‘RGB8toKCMYcm’:
cupsfilters/bitmap.c:417:3: warning: implicit declaration of function ‘cupsImageRGBToCMYK’ [-Wimplicit-function-declaration]
  417 |   cupsImageRGBToCMYK(src,cmyk,1);
      |   ^~~~~~~~~~~~~~~~~~
```

V1krant:
> I have merged it now, thank you very much.
> 
> Could you have a look at the following warning and fix it:
> 
> ```
> cupsfilters/bitmap.c: In function ‘RGB8toKCMYcm’:
> cupsfilters/bitmap.c:417:3: warning: implicit declaration of function ‘cupsImageRGBToCMYK’ [-Wimplicit-function-declaration]
>   417 |   cupsImageRGBToCMYK(src,cmyk,1);
>       |   ^~~~~~~~~~~~~~~~~~
> ```

Sorry Till, this was caused due to a mistake while updating the code. It can be fixed using the line `#include "image.h"
` in `cupsfilters/bitmap.c`. I would be making a pull request for the same soon.

tillkamppeter:
I have added the missubg `#include` now, thank you very much.

V1krant:
> I have added the missing `#include` now, thank you very much.

Thanks Till.

Issue #251, opened 2020-06-23
Can driverless list printers via _ipps._tcp


tangyanli:
Hi,

Is there some method to set driverless lists the network printers via _ipps._tcp?

PS:
I can use ippfind to list the network printers via _ipps._tcp as below.
$ ippfind _ipps._tcp

nidhijainpnp:
Hi @tangyanli ,


> PS:
> I can use ippfind to list the network printers via _ipps._tcp as below.
> $ ippfind _ipps._tcp

Yes,it is right that ,you can list the network printers and show their uri by `ippfind`.But,I am unable to understand your issue.Can you please elaborate more,what actually you want to do /or the error you are facing?

tillkamppeter:
The difference between `ippfind` and `driverless` is that the former lists all IPP printers, including remote CUPS queues and IPP printers which do not support driverless printing. `driverless` uses `ipptool` but calls it with filter options, exactly showing only printers which support at least IPP version 2.0 and at least one of the Page Description Languages (PDLs) PDF, Apple Raster, PWG Raster, and PCLm.
In addition, `ippfind` shows URIs in standard IPP format whereas `driverless` shows URIs in DNS-SD-service-name-based format.
@tangyanli suggests to add support for IPPS-based (encrypted) IPP print services.
My suggestion is, by default to show each printer once, if it appears with both IPP and IPPS, show the IPPS version. Also introduce command line options to show only IPP entries or only IPPS entries.

tangyanli:
@nidhijainpnp 
The issue I faced is: 
If I set IPP->TLS off at the printer side. I can find the printer at the web page: [localhost:631]->[Administration]->[Add Printer].
![1](https://user-images.githubusercontent.com/54799082/87491331-b9ed7980-c5fc-11ea-83bd-30724ebc4b27.png)
But if I set IPP->TLS on at the printer side. I can't find the printer at above web page any more.

@tillkamppeter Thanks for your explanation and suggestion. And I think your suggestion can properly resolved this issue.

tangyanli:
Thanks for your effort!

tangyanli:
Sometimes driverless will return error code 2. 
Is this normal situation?

![driverless-ng](https://user-images.githubusercontent.com/54799082/101717491-36717080-3a54-11eb-97da-72814958878e.png)

PR #270, opened 2020-07-29
changes in driverless.c file to list IPPS printers

nidhijainpnp:
Currently `driverless` list only IPP printers . Adding `_ipps._tcp` argument to driverless. 
The below command would list  the driver URIs and metadata for all available IPPS printers supporting driverless printing.
`driverless _ipps._tcp` 
Fixes [#251](https://github.com/OpenPrinting/cups-filters/issues/251)

EDIT: Now, `driverless` shows IPPS queues whenever the printer supports it and IPP queues when the printer does not support IPPS. 
 `driverless _ipps._tcp` shows only IPPS printer URI
`driverless _ipp._tcp` shows only IPP printer URI.

I have tested with `ippeveprinter` and it was giving results as expected.

### DIFF
--- a/utils/driverless.1
+++ b/utils/driverless.1
@@ -53,11 +53,11 @@ for driverless PPD file generation (to be used by CUPS).
 .TP
 .B
 \fB_ipps._tcp\fP
-List the Printer URIs for all available IPPS printers supporting driverless printing.
+Check for only IPPS printers supporting driverless printing (to be  used by CUPS).
 .TP
 .B
 \fB_ipp._tcp\fP
-List the Printer URIs for all available IPP printers supporting driverless printing.
+Check for only IPP printers supporting driverless printing (to be  used by CUPS).
 .TP
 .B
 \fBcat\fP \fIdriver URI\fP
--- a/utils/driverless.1
+++ b/utils/driverless.1
@@ -53,11 +53,11 @@ for driverless PPD file generation (to be used by CUPS).
 .TP
 .B
 \fB_ipps._tcp\fP
-Check for only IPPS printers supporting driverless printing (to be  used by CUPS).
+Check for only IPPS printers supporting driverless printing .
 .TP
 .B
 \fB_ipp._tcp\fP
-Check for only IPP printers supporting driverless printing (to be  used by CUPS).
+Check for only IPP printers supporting driverless printing .
 .TP
 .B
 \fBcat\fP \fIdriver URI\fP
--- a/utils/driverless.1
+++ b/utils/driverless.1
@@ -5,7 +5,7 @@
 .SH SYNOPSIS
 .nf
 .fam C
-\fBdriverless\fP [\fB-h\fP | \fB--help\fP | \fB--version\fP] [\fB-d\fP | \fB-v\fP | \fB--debug\fP] [\fBlist\fP] [\fB_ipps._tcp\fP] | [\fBcat\fP \fIdriver URI\fP] | [\fIIPP printer URI\fP]
+\fBdriverless\fP [\fB-h\fP | \fB--help\fP | \fB--version\fP] [\fB-d\fP | \fB-v\fP | \fB--debug\fP] [\fBlist\fP] [\fB_ipps._tcp\fP] [\fB_ipp._tcp\fP] | [\fBcat\fP \fIdriver URI\fP] | [\fIIPP printer URI\fP]
 
 .fam T
 .fi
@@ -48,13 +48,16 @@ Debug mode, verbose logging to stderr.
 .TP
 .B
 \fBlist\fP
-List the driver URIs and metadata for all available IPP printers suitable
+List the driver URIs and metadata for all available IPP/IPPS printers suitable
 for driverless PPD file generation (to be used by CUPS).
 .TP
 .B
 \fB_ipps._tcp\fP
-List the driver URIs and metadata for all available IPPS printers suitable
-for driverless PPD file generation (to be used by CUPS).
+List the Printer URIs for all available IPPS printers supporting driverless printing.
+.TP
+.B
+\fB_ipp._tcp\fP
+List the Printer URIs for all available IPP printers supporting driverless printing.
 .TP
 .B
 \fBcat\fP \fIdriver URI\fP
--- a/utils/driverless.1
+++ b/utils/driverless.1
@@ -5,7 +5,7 @@
 .SH SYNOPSIS
 .nf
 .fam C
-\fBdriverless\fP [\fB-h\fP | \fB--help\fP | \fB--version\fP] [\fB-d\fP | \fB-v\fP | \fB--debug\fP] [\fBlist\fP] | [\fBcat\fP \fIdriver URI\fP] | [\fIIPP printer URI\fP]
+\fBdriverless\fP [\fB-h\fP | \fB--help\fP | \fB--version\fP] [\fB-d\fP | \fB-v\fP | \fB--debug\fP] [\fBlist\fP] [\fB_ipps._tcp\fP] | [\fBcat\fP \fIdriver URI\fP] | [\fIIPP printer URI\fP]
 
 .fam T
 .fi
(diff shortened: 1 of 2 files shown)


tillkamppeter:
Thanks for this pull request, but this is not actually what I need.

It is useful for development and debugging, as you can find printers in the network which accept jobs encrypted via IPPS, but it does not improve the practical day-to-day use of "driverless" and so does not add to the user experience.
In my first e-mail, to introduce you into your work, in the section "3. Fax support" I have explained you how the "driverless" utility is used by CUPS to discover driverless IPP printers and to easily allow to set up print queues for them with any printer setup tool. The expected user experience is now that each physical printer is listed exactly once,  NOT two times if it supports both IPP and IPPS, and the one listing of the printer uses the more sophisticated protocol, meaning IPPS if both are supported. So all available IPP printers get listed, the ones which support ONLY IPP and NOT IPPS are listed as IPP, the ones which support IPPS are ALWAYS listed as IPPS, independent whether they also support IPP. PPD generation should always work, independent if the IPP URI used is IPP or IPPS (as long as the printer supports it). This way CUPS (or the printer setup tool) should create queues with the "better" protocol.

So your task here is to modify "driverless" so that when used with CUPS you get IPPS queues whenever the printer supports it and IPP queues when the printer does not support IPPS. Please test this with CUPS.

For testing most physical printers support IPPS, the sample Printer Application of [PAPPL](https://github.com/michaelrsweet/pappl/) and the [PCL example Printer Application](https://github.com/michaelrsweet/hp-printer-app/) also. IPP-over-USB with [ipp-usb](https://github.com/OpenPrinting/ipp-usb) does not support IPPS.

tillkamppeter:
Thank you very much for the new patch, but there are several things which I do not like:
1. Do not use temporary files for the two printer lists. One can easily drop them in some array (you are reading them into arrays anyway, why not dropping the items into arrays directly?)
2. Temporary files should never have static (predictable) names for security reasons and they should be dropped into some temporary file directory (when solving (1) this item gets irrelevant anyway)
3.  Do not use arrays with static amount of items (101 here). To create dynamic arrays you do not need to re-invent the wheel by using the CUPS arrays of libcups (see `cups/array.h` in the source code of CUPS). We are using libcups anyway in cups-filters, so no problem to use the CUPS arrays. With CUPS arrays you can even easily sort the output, find and reject duplicates, ... (with a well-selected comparison function you could easily reject the IPP version of a printer if the IPPS version is already there).

Could you fix these points?

nidhijainpnp:
Sure,but I want to clarify following points:
For first part(1)-  In `int list_printers()` function ,I have to read the output  from child process and if I store the output( printer uris) in some array,it would not get saved as after the exit of child process,any changes made to that array would be lost(any memory allocated to child process would be destroyed).Is there any way by which I could restore the changes made to that array?How could I then drop printer list in some array so that changes won't be lost?
and second thing is I am copying only service_name in arrays not service_uri because if we run `driverless list` ,it would list the metadata also ,because of which I would have to allocate about 10^4 Bytes  of memory(calculated based on the varibale size of service_uri,service_hostname ...declared) for single printer.Would it be feasible to allocate this much memory?

tillkamppeter:
You can already see in the original `utils/driverless.c` how the output of `ippfind` is caught. driverless forks into two processes, one runs `ippfind` and directs the output into stdout, the other grabs this output via stdin. The two processes are connected in a way that stdout of the first process is piped to stdin of the second process. So one process is running `ippfind` while the other is processing the output. This second process can, instead of sending its results finally to stdout, store them in some variable/array.
You will have a main process, which forks out an auxiliary process. The auxiliary process runs `ippfind` to get the IPPS entries, the main process receives `ippfind`'s output via pipe. The results of processing it stores in an array. When `ippfind` finishes, the auxiliary process terminates. The main process stays running and forks out an auxiliary process for a second time, this time letting the auxiliary process run `ippfind` for the IPP entries. The main process again processes the output and adds the results to the array, but only the ones which do not already exist from the IPP round. It depends on the user-requested tasks which `ippfind` calls are actually done.

utils/driverless.1
```diff
@@ -48,10 +48,18 @@ Debug mode, verbose logging to stderr.
...
+List the driver URIs and metadata for all available IPP/IPPS printers suitable
 for driverless PPD file generation (to be used by CUPS).
 .TP
 .B
+\fB_ipps._tcp\fP
+Check for only IPPS printers supporting driverless printing (to be  used by CUPS).
+.TP
+.B
+\fB_ipp._tcp\fP
+Check for only IPP printers supporting driverless printing (to be  used by CUPS).
```
tillkamppeter, line 60: Note that these command line arguments are only used for the manual call of `driverless` from a terminal windows. The remark "(to be used by CUPS)" in the description of the argument means that the argument is used by CUPS, like the `list` and `cat` arguments. So could you please remove the "(to be used by CUPS)" for these two arguments.

nidhijainpnp: Sure, I would made the respective changes in driverless.c file also ,to use these command  line arguments  only for the manual call of `driverless ` from a terminal windows.

tillkamppeter: You mean in the help output? Yes, please remove it there, too.

tillkamppeter:
I have merged your changes. Thank you very much.

When testing I found the following issues:

Please let `driverless` not issue the CUPS log messages (`DEBUG: ...`) when calling in a terminal. These messages should only be issued if CUPS is calling the program.

I have the following URIs available:
```
ipps://HP%20OfficeJet%20Pro%208730%20%5B08C229%5D._ipps._tcp.local/
ipp://HP%20OfficeJet%20Pro%208730%20%5B08C229%5D._ipp._tcp.local/
ipp://HP%20OfficeJet%20Pro%208730%20%5B08C229%5D%20(USB)._ipp._tcp.local/
```
In this case `driverless` (without arguments) should show
```
ipps://HP%20OfficeJet%20Pro%208730%20%5B08C229%5D._ipps._tcp.local/
ipp://HP%20OfficeJet%20Pro%208730%20%5B08C229%5D%20(USB)._ipp._tcp.local/
```
as the second in the above list is the IPP version of the first. What it actually shows is only
```
ipps://HP%20OfficeJet%20Pro%208730%20%5B08C229%5D._ipps._tcp.local/
```

nidhijainpnp:
> Please let `driverless` not issue the CUPS log messages (`DEBUG: ...`) when calling in a terminal. These messages should only be issued if CUPS is calling the program.

I would set `debug = 0` for `driverless` in my next pull-request. 

> as the second in the above list is the IPP version of the first. What it actually shows is only
> 
> ```
> ipps://HP%20OfficeJet%20Pro%208730%20%5B08C229%5D._ipps._tcp.local/
> ```
Could you please also test and confirm me if `driverless  _ipps._tcp ` and `driverless _ipp._tcp` gave the expected results or not. (This would help me for better debugging until I got my printer).

tillkamppeter:
Yes, set `debug = 0` for manual calls, like `driverless` without arguments and not detecting running CUPS by SOFTWARE env variable.

On my system I get (DEBUG: messages removed):
```
$ ippfind _ipps._tcp
ipps://HP18602408C229.local:443/ipp/print
ipps://till-x1yoga.local:631/printers/snap-ps
$ ippfind
ipp://HP18602408C229.local:631/ipp/print
ipp://till-x1yoga.local:60000/ipp/print
ipp://till-x1yoga.local:631/printers/snap-ps
$ ./driverless _ipps._tcp
ipps://HP%20OfficeJet%20Pro%208730%20%5B08C229%5D._ipps._tcp.local/
$ ./driverless _ipp._tcp
ipp://HP%20OfficeJet%20Pro%208730%20%5B08C229%5D._ipp._tcp.local/
ipp://HP%20OfficeJet%20Pro%208730%20%5B08C229%5D%20(USB)._ipp._tcp.local/
$ ./driverless
ipps://HP%20OfficeJet%20Pro%208730%20%5B08C229%5D._ipps._tcp.local/
$ 
```
The URI with `(USB)` is the IPP-over-USB version of my printer, provided via ipp-usb. `ippfind` shows it as `ipp://till-x1yoga.local:60000/ipp/print`.
The URI `ipp(s)://till-x1yoga.local:631/printers/snap-ps` which shows only in the `ippfind` output is a CUPS queue which is intendedly filtered out by `driverless`.

PR #275, opened 2020-08-09
Convert pclmtoraster to a filter function

V1krant:


tillkamppeter:
Generally looks OK for me, but please make sure the following:
- A Printer Application linking libcupsfilters can run several jobs in parallel, as separate threads or forked sub processes. This means that the filter functions must work in parallel, being thread-safe. So please check whether this is OK with global variables and/or `namespace {...}` which you are using.
- Does everything build without warnings?

V1krant:
> - Does everything build without warnings?

@tillkamppeter, currently everything build perfectly for me without any warnings.

> - A Printer Application linking libcupsfilters can run several jobs in parallel, as separate threads or forked sub processes. This means that the filter functions must work in parallel, being thread-safe. So please check whether this is OK with global variables and/or `namespace {...}` which you are using.

Sure. I’ll check this and update the code accordingly if necessary.

tillkamppeter:
Could you also rename `filter/pclmtoraster.cxx` to `filter/pclmtoraster.c`. This file does not contain any C++ any more.

PR #280, opened 2020-08-16
Add fax support

nidhijainpnp:


nidhijainpnp:
Testing Results:
```
nidhi@nidhi-X510UNR:~$ driverless-fax
DEBUG: driverless-fax argv[0] = 
ipps://HP%20OfficeJet%20Pro%206960%20%5BC3B82D%5D._ipps._tcp.local/
ipp://HP%20OfficeJet%20Pro%206960%20%5BC3B82D%5D%20(USB%201)._ipp._tcp.local/

nidhi@nidhi-X510UNR:~$ driverless-fax --std-ipp-uris
DEBUG: driverless-fax argv[1] = --std-ipp-uris
ipps://HP30E171C3B82D.local:443/ipp/faxout
ipp://HP30E171C3B82D.local:631/ipp/faxout
ipp://nidhi-X510UNR.local:60000/ipp/faxout

nidhi@nidhi-X510UNR:~$ driverless-fax _ipps._tcp
DEBUG: driverless-fax argv[1] = _ipps._tcp
ipps://HP%20OfficeJet%20Pro%206960%20%5BC3B82D%5D._ipps._tcp.local/

nidhi@nidhi-X510UNR:~$ driverless-fax _ipp._tcp
DEBUG: driverless-fax argv[1] = _ipp._tcp
ipp://HP%20OfficeJet%20Pro%206960%20%5BC3B82D%5D._ipp._tcp.local/
ipp://HP%20OfficeJet%20Pro%206960%20%5BC3B82D%5D%20(USB%201)._ipp._tcp.local/

nidhi@nidhi-X510UNR:~$ driverless-fax list
DEBUG: driverless-fax argv[1] = list
DEBUG: Started ippfind (PID 14369)
"driverless-fax:ipps://HP%20OfficeJet%20Pro%206960%20%5BC3B82D%5D._ipps._tcp.local/" en "HP" "HP OfficeJet Pro 6960, fully driverless, cups-filters 1.27.5" "MFG:HP;MDL:OfficeJet Pro 6960;CMD:PCLM,PCL,PWGRaster,AppleRaster,JPEG,URF,PWG;"
DEBUG: Started ippfind (PID 14371)
"driverless-fax:ipp://HP%20OfficeJet%20Pro%206960%20%5BC3B82D%5D._ipp._tcp.local/" en "HP" "HP OfficeJet Pro 6960, fully driverless, cups-filters 1.27.5" "MFG:HP;MDL:OfficeJet Pro 6960;CMD:PCLM,PCL,PWGRaster,AppleRaster,JPEG,URF,PWG;"
"driverless-fax:ipp://HP%20OfficeJet%20Pro%206960%20%5BC3B82D%5D%20(USB%201)._ipp._tcp.local/" en "HP" "HP OfficeJet Pro 6960, fully driverless, cups-filters 1.27.5" "MFG:HP;MDL:OfficeJet Pro 6960;CMD:PCLM,PCL,PWGRaster,AppleRaster,JPEG,URF,PWG;"
DEBUG: PID 14369 (ippfind _ipps._tcp) exited with no errors.
DEBUG: PID 14371 (ippfind _ipp._tcp) exited with no errors.
```
Implemented `ippfind_based_uri_converter()` .It is about 0.02 sec faster than the cupsBackendDeviceURI()
function, hence,used it for both printer and fax URIs.
This command successfully generates the ppd.
```
driverless-fax cat driverless-fax:ipps://HP%20OfficeJet%20Pro%206960%20%5BC3B82D%5D._ipps._tcp.local/
```

tillkamppeter:
Some issues:
1. IPP URIs do not always end with `/ipp/print` and fax URIs not always with `/ipp/faxout`. Especially Printer Applications can support more than one physical device or high-end devices could have more than one print engine. See [here](https://openprinting.github.io/OpenPrinting-News-March-2020/#printer-application-frameworksdk---pappl) for examples. Therefore you always have to insert the actual content of the "rp" and "rfo" TXT record fields. You can read out these fields with `ippfind`, with the same command line as you are already using for reading out the service name and the port.
2. Tou have replaced the use of the function cupsBackendDeviceURI() by your new ippfind_based_uri_converter() function, but put the new function into `utils/driverless.c` and not into `cupsfilters/ipp.c`. You also removed the call of the old function from the get-printer-attributes functions, which probably breaks `cups-browsed` which also uses this function. You should put ippfind_based_uri_converter() into `cupsfilters/ipp.c` and perhaps allow some switch between the two. get-printer-attributes needs a URI converter function call for `cups-browsed` to work.
Also how did you find out that your new function is 0.02 seconds faster? And what is the speed advantage in percent?

nidhijainpnp:
>You should put ippfind_based_uri_converter() into cupsfilters/ipp.c and perhaps allow some switch between the two. get-printer-attributes needs a URI converter function call for cups-browsed to work.

I have to change the defintion of `get-printer-attributes` function to add one more argument to check if its getting called by `cups-browsed` or `driverless`(as only these two calls the `get-printer-attributes`)and therfore could switch between `ippfind_based_uri_converter()` and `cupsBackendDeviceURI() ` accordingly. Would it be fine?

>Also how did you find out that your new function is 0.02 seconds faster?

I have used clock() function available in time.h

tillkamppeter:
To add a parameter to a function, changing its call scheme, do not change it straight away, as someone could use libcupsfilters for something which we do not know about. Therefore one has to keep the API. You see already that we have get_printer_attributes(), get_printer_attributes2(), and get_printer_attributes3(), with slightly different call schemes. During the time since we introduced the function we needed some call scheme alternatives.
In (the original) driverless.c I see a line
```
response = get_printer_attributes(uri, NULL, 0, NULL, 0, 1);
```
It is a call of the get_printer_attributes() function which takes 6 arguments. Assume you want to change this one to add an is_fax argument:
```
response = get_printer_attributes4(uri, NULL, 0, NULL, 0, 1, is_fax);
```
You see that we introduce a new name, to not change the old name's call scheme. Now we need to modify ipp.c (and ipp.h) to have the get_printer_attributes4() which takes the extra, boolean argument for whether the fax channel arguments of the device specified by `uri` should be polled.
```
What you have to do now is to have a look into ipp.c which function does the actual "dirty" work. It is get_printer_attributes3(). Copy the
```
ipp_t *
get_printer_attributes3(http_t *http_printer,
			const char* raw_uri,
			const char* const pattrs[],
			int pattrs_size,
			const char* const req_attrs[],
			int req_attrs_size,
			int debug,
                        int* driverless_info)
{
```
To right over the comment above it. Then, at the original, add an extra `, int is_fax` at the end and change the name from `get_printer_attributes3` to `get_printer_attributes5`. To the copy above you add
```
  get_printer_attributes5(http_printer, raw_uri, pattrs[], pattrs_size, req_attrs[], req_attrs_size, debug, driverless_info, 0);
}
```
to complete the function definition. So the new get_printer_attributes3() calls get_printer_attributes5() with `is_fax` set to `0`, so that it behaves as the old function. The new function get_printer_attributes5() which now does all the dirty work you need to modify, so that it resolves the URI with your new function (which you have to move into ipp.c). Your new function simply takes the `is_fax` argument and so does the right thing.
Now add a get_printer_attributes4() with the same call scheme as get_printer_attributes() but with `is_fax` added. This one calls get_printer_attributes5() then.

File: utils/driverless.c [hunk 1]
```diff
@@ -673,19 +679,221 @@ list_printers (int mode ,int reg_type_no)
...
+    fd, 
+    bytes,
+    exit_status_uri_converter = 0;
+  char *ptr1, 
+       *ptr2,
+       *std_uri = NULL;
+   /* Request printer properties via IPP to generate a PPD file for the
      printer */
-  response = get_printer_attributes(uri, NULL, 0, NULL, 0, 1);
+  std_uri = (char *)malloc(2048*(sizeof(char)));
```
zdohnal, line 889: Hi @nidhijainpnp ,

please free any heap allocated memory before you exit the function if you don't return it as a parameter/return value.

cupsfilters/ipp.c
```diff
@@ -333,4 +418,224 @@ get_printer_attributes3(http_t *http_printer,
   if (have_http == 0) httpClose(http_printer);
   return NULL;
 }
+const char*
+ippfind_based_uri_converter (const char *uri ){
```
tillkamppeter, line 422: Why don't you keep the "is_fax" switch here. As you are putting this function into the API (the ipp.h file) it should be universal for both print and fax, especially with its name ippfind_based_uri_converter().
Make it
```
const char*
￼ippfind_based_uri_converter (const char *uri, int is_fax){
```
and inside switch between usinf the "rp" or the "rfo" TXT field for the resource.

tillkamppeter:
Also make sure that cups-browsed stays working.

cupsfilters/ipp.c
```diff
@@ -262,10 +262,10 @@ get_printer_attributes5(http_t *http_printer,
   get_printer_attributes_log[0] = '\0';
 
   /* Convert DNS-SD-service-name-based URIs to host-name-based URIs */
-  if(!isFax)
+  if(is_fax == -1)
```
tillkamppeter, line 265: Is it not easier to have this variable set to 0 for requesting the printing URI and 1 for requesting the fax URI (instead of -1 and 0)?

cupsfilters/ipp.c
```diff
@@ -262,10 +262,10 @@ get_printer_attributes5(http_t *http_printer,
...
     uri = resolve_uri(raw_uri);
```
tillkamppeter, line 266: Why do you use resolve_uri() for print URIs and ippfind_based_uri_converter() for fax URIs? You could use ippfind_based_uri_converter() for both. Or is resolve_uri() faster due to the fact that it does not call an external executable?
Another possibility would be to make the caller somehow select which of the two functions is used.
Currently cups-browsed could use resolve_uri()  and driverless could use ippfind_based_uri_converter(), but later you will also add fax support to cups-browsed.
If there is actually a performance penalty by calling the external `ippfind` executable then it would be best to do all print URI conversions with resolve_uri() for performance and all fax URI conversions with ippfind_based_uri_converter() for fax support.

nidhijainpnp: >Why do you use resolve_uri() for print URIs and ippfind_based_uri_converter() for fax URIs? You could use ippfind_based_uri_converter() for both. Or is resolve_uri() faster due to the fact that it does not call an external executable?

In `utils/driverless.c` file ,`isFax` is set to "1" if we run commands with `driverless-fax` else `isFax` is set to "0"(for only `driverless` calls).So ,`ippfind_based_uri_converter()` is for both - print uris and as well fax uris. `is_fax` is set to "-1" (while calling `get_printer_attributes5()` from `get_printer_attributes3()`) for calls from `cups-browsed`, so that `cups-browsed` use `resolve_uri`. 
>If there is actually a performance penalty by calling the external ippfind executable then it would be best to do all print URI conversions with resolve_uri() for performance and all fax URI conversions with ippfind_based_uri_converter() for fax support.
 
I have used [clock()](http://www.gnu.org/software/libc/manual/html_node/CPU-Time.html) function available in time.h (or is there any better way to check?)to performance check . But both `ippfind_based_uri_converter()` and `resolve_uri()` were giving approximately the same time (~0.026 sec).

utils/driverless.c
tillkamppeter: @nidhijainpnp, did you check whether all allocated memory gets correctly freed?

tillkamppeter:
Thanks, @nidhijainpnp, now you only need to check whether all allocated memory (which is not a return value) gets freed before exiting the function, and if allocated memory is a return value, that it can be freed by the caller (the pointer returned has to point at the beginning of the allocated memory).

utils/driverless.c
nidhijainpnp: yes

nidhijainpnp:
Till,
I have checked ,all allocated memory gets freed . Allocated memory by `ippfind_based_uri_converter()`in cupsfilters/ipp.c is returned to `generate_ppd()` in utils/driverless.c as `ipp_t response` which is freed by `ippDelete(response)`

tillkamppeter:
Great, thanks.
Now we need the second part. The PPD file generator needs fax support. The PPD file generator is in the file `cupsfilters/ppdgenerator.c`. For a fax queue the PPD has to get two items added:
First, the PPD must tell CUPS that it is for fax. For this, it needs
```
*cupsIPPFaxOut: True
```
This makes the function cupsBackendDeviceURI() of libcups (in `cups/backend.c` of CUPS source code) identify a queue as fax queue and so resolves the DNS-SD-based URI to an IPP-based fax URI.
second, we have to tell the print dialog (at least it should work with the GTK one) that we want to have an option named "phone" which takes a string. This is the phone number to be dialed by the printer where the fax should go to.
For this we want to add something like
```
*% Base "Phone" option
*OpenUI Phone/Phone Number: PickOne
*OrderDependency: 10 AnySetup *Phone
*DefaultPhone: None
*Phone None: ""
*CloseUI: *Phone
*% Custom "Phone" option
*CustomPhone True: ""
*ParamCustomPhone Text: 1 string 0 64
```
to the PPD file. This adds an option named "Phone" and using CUPS' PPD extensions the print dialog is told to show a string input field for a maximum of 64 characters. The user can then enter the phone number here or copy and paste it from his contact list.
The option entry is always the same, regardless of printer. The "Phone" is a standard IPP option understood by all IPP printers with fax support.
Could you add your changes in the PPD generator to this Pull Request? Thanks.

nidhijainpnp:
I have read this from the standard IPP Fax Out Standard Abstract of June 18, 2014:
```
ipp-features-supported (1setOf type2 keyword)
This specification defines the REQUIRED keyword 'faxout' for the "ipp-features-supported" Printer attribute.
```
 It is given in the table of  REQUIRED Printer Description attributes for a Printer. So,I have checked the fax queue with `ipp-features-supported` attribute as key and `faxout` as value.

tillkamppeter:
> I have read this from the standard IPP Fax Out Standard Abstract of June 18, 2014:
> 
> ```
> ipp-features-supported (1setOf type2 keyword)
> This specification defines the REQUIRED keyword 'faxout' for the "ipp-features-supported" Printer attribute.
> ```
> 
> It is given in the table of REQUIRED Printer Description attributes for a Printer. So,I have checked the fax queue with `ipp-features-supported` attribute as key and `faxout` as value.

OK, great, then we go this way, as you did.

nidhijainpnp:
>You can always test with CUPS' cupstestppd utility.

I have checked with both `phone`(lowercase) and `Phone`(uppercase). Both ways were working. But ,I have used lowercase one as you have suggested.
>print a fax via lp -d fax-printer -o phone=12345678 file.pdf, ...

I have also checked this ,it is successfully printing fax.
`lp -d HP_OfficeJet_Pro_6960_C3B82D_ -o phone=12345678 /home/nidhi/Desktop/finally.pdf`

tillkamppeter:
Thank you very much. I have merged it now and also will pass it over to 1.x branch.

PR #282, opened 2020-08-18
Introduced rastertops filter function

JaiLuthra1:


cupsfilters/rastertops.c
```diff
@@ -0,0 +1,527 @@
...
+    if (doc->logfunc) doc->logfunc(doc->logdata, FILTER_LOGLEVEL_ERROR,
+		  "invalid compression level\n");
+    break;
+  case Z_DATA_ERROR:
+    if (doc->logfunc) doc->logfunc(doc->logdata, FILTER_LOGLEVEL_ERROR,
+		  "invalid or incomplete deflate data\n");
+    break;
+  case Z_MEM_ERROR:
+    if (doc->logfunc) doc->logfunc(doc->logdata, FILTER_LOGLEVEL_ERROR,
+		  "out of memory\n");
```
tillkamppeter, line 327: When logging, please start the log messages with "rastertops: ", for example here do
```
if (doc->logfunc) doc->logfunc(doc->logdata, FILTER_LOGLEVEL_ERROR,
￼		  "rastertops: out of memory\n");
```
This makes it easier to see from which filter function a log message comes, especially when calling a chain of filters.
See also the other filter functions.
Could you correct this with all log messages? Thanks.

tillkamppeter (approved):

PR #281, opened 2020-08-18
cups-browsed.c: Fix several memory leaks

zdohnal:
Hi,

we got an [issue in Fedora](https://bugzilla.redhat.com/show_bug.cgi?id=1867412) - cups-browsed accumulates memory when it is run for long time (up to 1.3GB after a day of running). 
[This memory report](https://github.com/OpenPrinting/cups-filters/files/5088515/cups2.log) was collected by attaching to the process by gdb and leak check was run.

The log showed several CUPS arrays weren't freed at all (or its pointer was rewritten by another allocation in next loop iteration), several non-freed char arrays and non-freed ```ipp_t``` structures.

The pull request was tested by reporter and he didn't report any significant leakage after a day.

Would you mind adding the pull request to the project?

Thank you in advance,

Zdenek

PR #283, opened 2020-08-19
cups-browsed.c: fix compiler warnings

zdohnal:
See #281


tillkamppeter:
Thank you very much, it is all working correctly now.

PR #292, opened 2020-08-26
[branch 1.x] driverless-fax: don't use a fixed path

zdohnal:
Hi,

```driverless-fax``` wrapper uses a fixed path to ```driverless```, which is a problem on distros which don't put CUPS stuff into ```/usr/lib/cups``` (IIRC archlinux puts it into ```/usr/libexec/cups```).
The patch renames the wrapper to {name}.in and uses ```@CUPS_SERVERBIN@``` macro, which will be substituted during compilation for the real path.

Would you mind adding the patch? I'll create an extra pull request for master too.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -22,6 +22,7 @@ EXTRA_DIST = \
 	libfontembed.pc.in \
 	utils/cups-browsed.service \
 	utils/cups-browsed-upstart.conf \
+	utils/driverless-fax.in \
 	filter/braille/drivers/index/ubrlto4dot.c \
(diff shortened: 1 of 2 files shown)


tillkamppeter:
Thank you for the fix, but could you make a PR to the master branch. All new development gets into the master branch and is cherry-piced to 1.x if it also applies to there.

zdohnal:
@tillkamppeter 
> I'll create an extra pull request for master too. 

:)

tillkamppeter:
Cherry-picked from master.

PR #301, opened 2020-08-29
Convert pdftopdf to a filter function

JaiLuthra1:


tillkamppeter:
Jai, thank you very much.

PR #303, opened 2020-08-31
braille: Support new liblouis tables not containing a display name

sthibaul:


tillkamppeter:
Thank you very much. I have merged it now and also added it to the 1.x branch, for the 1.28.2 release.

PR #305, opened 2020-09-05
driverless.c: creating two more forks to read output of each ippfind call simultaneously

nidhijainpnp:
time taken after creating two more forks to read output of each ippfind call simultaneously
```
nidhi@nidhi-X510UNR:~$ time driverless
ipps://hello._ipps._tcp.local/
ipps://hello1._ipps._tcp.local/
ipps://hello2._ipps._tcp.local/

real	0m3.224s
user	0m0.029s
sys	0m0.027s

nidhi@nidhi-X510UNR:~$ time driverless list
DEBUG: Started ippfind (PID 18725)
DEBUG: Started ippfind (PID 18726)
DEBUG: Started reading ipp data (PID 18727)
DEBUG: Started reading ipps data (PID 18728)
"driverless:ipps://hello._ipps._tcp.local/" en "Example" "Example Printer, fully driverless, cups-filters 1.27.5" "MFG:Example;MDL:Printer;CMD:PWGRaster,AppleRaster,PWG,URF;"
"driverless:ipps://hello1._ipps._tcp.local/" en "Example" "Example Printer, fully driverless, cups-filters 1.27.5" "MFG:Example;MDL:Printer;CMD:PWGRaster,AppleRaster,PWG,URF;"
"driverless:ipps://hello2._ipps._tcp.local/" en "Example" "Example Printer, fully driverless, cups-filters 1.27.5" "MFG:Example;MDL:Printer;CMD:PWGRaster,AppleRaster,PWG,URF;"
DEBUG: PID 18725 (ippfind _ipps._tcp) exited with no errors.
DEBUG: PID 18726 (ippfind _ipp._tcp) exited with no errors.
DEBUG: PID 18727 (read ipp data) exited with no errors.
DEBUG: PID 18728 (read ipps data) exited with no errors.

real	0m3.069s
user	0m0.037s
sys	0m0.030s
```
time taken previously when one ippfind calls runs one after the other
(removed the output lines and only time is showing here)
```
nidhi@nidhi-X510UNR:~$ time driverless
real	0m4.274s
user	0m0.023s
sys	0m0.028s

nidhi@nidhi-X510UNR:~$ time driverless list
real	0m3.134s
user	0m0.034s
sys	0m0.067s
```

### DIFF
--- a/utils/driverless.c
+++ b/utils/driverless.c
@@ -58,14 +58,104 @@ convert_to_port(char *a)
 
   return (port);
 }
+/*
+ * 'copy_output()' - copy ippfind output data .
+ */
+
+static int				/* O - 0 on success, -1 on error */
+copy_output(int read_fd,		/* I - Print file descriptor */
+             int write_fd)		/* I - Device file descriptor */
+{
+  int		nfds;			/* Maximum file descriptor value + 1 */
+  fd_set	input;			/* Input set for reading */
+  ssize_t	print_bytes,		/* Print bytes read */
+		bytes;			/* Bytes written */
+  char		print_buffer[MAX_OUTPUT_LEN],	/* Print data buffer */
+		*print_ptr;		/* Pointer into print data buffer */
+  struct timeval timeout;		/* Timeout for read... */
+
+
+ /*
+  * Figure out the maximum file descriptor value to use with select()...
+  */
+  nfds = (read_fd > write_fd ? read_fd : write_fd) + 1;
+
+ /*
+  * Now loop until we are out of data from read_fd...
+  */
+
+  for (;;)
+  {
+   /*
+    * Use select() to determine whether we have data to copy around...
+    */
+
+    FD_ZERO(&input);
+    FD_SET(read_fd, &input);
+    timeout.tv_sec  = 2;
+    timeout.tv_usec = 0;
+
+    if (select(nfds, &input, NULL, NULL, &timeout) < 0)
+      return (-1);
+    if (!FD_ISSET(read_fd, &input))
+     {
+      return (0);
+     } 
+
+    if ((print_bytes = read(read_fd, print_buffer,
+			    sizeof(print_buffer))) < 0)
+    {
+     /*
+      * Read error - bail if we don't see EAGAIN or EINTR...
+      */
+
+      if (errno != EAGAIN && errno != EINTR)
+      {
+        perror("ERROR: Unable to read print data");
+	      return (-1);
+      }
+
+      print_bytes = 0;
+    }
+    else if (print_bytes == 0)
+    {
+     /*
+      * End of file, return...
+      */
+      
+      return (0);
+    }
+    
+    for (print_ptr = print_buffer; print_bytes > 0;)
+    {
+      if ((bytes = write(write_fd, print_ptr, print_bytes)) < 0)
+      {
+       /*
+        * Write error - bail if we don't see an error we can retry...
+	    */
+
+        if (errno != ENOSPC && errno != ENXIO && errno != EAGAIN &&
+	       errno != EINTR && errno != ENOTTY)
+          {
+            perror("ERROR: Unable to write print data");
+            return (-1);
+          }
+      }
+      else
+      {
+        print_bytes -= bytes;
+	      print_ptr   += bytes;
+      }
+    }
+  }
+}
 
 void 
 listPrintersInArray(int reg_type_no, int mode, int isFax, char* ippfind_output ) {
   int	driverless_support = 0, /*process id for ippfind */
         port,
         is_local;
-        
-
+  
   char	buffer[8192],		/* Copy buffer */
         *ptr,		        /* Pointer into string */
         *scheme = NULL,
@@ -543,14 +633,87 @@ list_printers (int mode, int reg_type_no, int isFax)
       fprintf(stderr, "DEBUG: Started %s (PID %d)\n", ippfind_argv[0],
 	      ippfind_ipp_pid);
   }
+
+
+  /*
+    Copy data to pipe
+  */
+  int read_data_pipe_ipp[2],
+      read_data_pipe_ipps[2];
+  int read_data_pid_ipp = 0,
+      read_data_pid_ipps = 0;
+  int read_status_ipp ,
+      read_status_ipps;
+
+  if (pipe(read_data_pipe_ipp)) {
+    perror("ERROR: Unable to create pipe to copy data for ipp ");
+    exit_status = 1;
+    goto error;
+  }
+  if (reg_type_no <= 1) {
+    if ((read_data_pid_ipp = fork()) == 0) 
+    {
+     /*
+      * Child comes here...
+      */
+      close(post_proc_pipe_ipp[1]);
+      close(read_data_pipe_ipp[0]);
+      read_status_ipp = copy_output(post_proc_pipe_ipp[0],read_data_pipe_ipp[1]);
+      if(read_status_ipp < 0)
+        exit(1);
+      exit(0);
+    } else if (read_data_pid_ipp < 0) {
+     /*
+      * Unable to fork!
+      */
+      perror("ERROR: Unable to read ipp data");
+      exit_status = 1;
+      goto error;
+    }
+    if (debug)
+      fprintf(stderr, "DEBUG: Started reading ipp data (PID %d)\n",
+	      read_data_pid_ipp);
+  }
+  
+
+  if (pipe(read_data_pipe_ipps)) {
+    perror("ERROR: Unable to create pipe to copy data for ipps ");
+    exit_status = 1;
+    goto error;
+  }
+  if (reg_type_no >= 1) {
+    if ((read_data_pid_ipps = fork()) == 0) 
+    {
+     /*
+      * Child comes here...
+      */
+      close(post_proc_pipe_ipps[1]);
+      close(read_data_pipe_ipps[0]);
+      read_status_ipps = copy_output(post_proc_pipe_ipps[0],read_data_pipe_ipps[1]);
+      if(read_status_ipps < 0)
+        exit(1);
+      exit(0);
+    } else if (read_data_pid_ipps < 0) {
+     /*
+      * Unable to fork!
+      */
+      perror("ERROR: Unable to read ipps data");
+      exit_status = 1;
+      goto error;
+    }
+    if (debug)
+      fprintf(stderr, "DEBUG: Started reading ipps data (PID %d)\n",
+	      read_data_pid_ipps);
+  }
+  
   /*
   Reading the ippfind output in CUPS Array
   */
   if(reg_type_no >=1)
   {
-    dup2(post_proc_pipe_ipps[0], 0);
-    close(post_proc_pipe_ipps[0]);
-    close(post_proc_pipe_ipps[1]);
+    dup2(read_data_pipe_ipps[0], 0);
+    close(read_data_pipe_ipps[0]);
+    close(read_data_pipe_ipps[1]);
 
     fp = cupsFileStdin();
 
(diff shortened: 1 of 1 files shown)


tillkamppeter:
Thanks for working on it, but it seems not to be the right way yet. Your measurement results are not much better than with the original result.
I have looked into your code. You have added the extra forks which I mentioned, but they do the wrong thing. They pick up the output of the `ippfind` calls and pass them on to the main process. The main process uses collects the data as before, first from the first `ippfind` call and then from the second. So the unfinished first `ippfind` is still blocking the second when it arrives at the first output-producing event.
What you have to do is to create two CUPS arrays for the output lines, one for the IPP `ippfind` call and one for the IPPS `ippfind` call. Now instead of letting your two new forks passing the read data on to the main process, put the lined into the CUPS array and send nothing to the main process.
As sson as the two `ippfind` calls end, and with them also the two new data collecting processes, the main process continues its work producing the final output from the two arrays containing the output lines of the two `ippfind` calls.
Could you do that?

tillkamppeter:
And by the way, you do not need select() for waiting for data on a single file descriptor. The select() on a single file descriptor blocks the program until data arrives on this file descriptor. If you read from the file descriptor right away it also blocks until data arrives and then the read happens, which should be the same thing. select() is for if you have more than one descriptor and do not know which one receives data first and you simply want to wait data from any of the descriptors and when there is data, read it. It is a very simple event loop. AFAIR CUPS uses this kind of event loop.

nidhijainpnp:
>What you have to do is to create two CUPS arrays for the output lines, one for the IPP ippfind call and one for the IPPS ippfind call. Now instead of letting your two new forks passing the read data on to the main process, put the lined into the CUPS array and send nothing to the main process.
> As sson as the two `ippfind` calls end, and with them also the two new data collecting processes, the main process continues its work producing the final output from the two arrays containing the output lines of the two `ippfind` calls.

But, how could I access the two arrays (which are in child process) to main process(parent process)? AFAIK ,  communication between parent and child process could be only done through pipes or shared memory concept.(But this would again be of no use due to time overhead).

tillkamppeter:
There are two possible solutions:
1. The two new processes which collect the output of the `ippfind` calls do not pass on the data to the main process right away. Instead they collect the data into a CUPS array as I told. Then in the end when the the corresponding `ippfind` process has ended (and so EOF is read by the collecting process), we send out the whole array content to the main process at once. The main process waits for both `ippfind` processes to end and then picks up the data from the collector processes, first from IPPS, after that from IPP.
2. You do not fork out collector processes, but have only the two `ippfind` subprocesses and the mainprocess. While the `ippfind` processes are running, the main process collects the output from both at once, using the select() or poll() function (now we have two file descriptors to take care of). To the select() or poll() functions you have to supply a list of file descriptors you want to take care of. You supply the output (read) ends of the two pipes which pass the output of the two `ippfind` processes to the main process. select() or poll() stops when one of the `ippfind` processes produces an output line, then the corresponding pipe has data to be read, select() or poll() tells you which one and you read the output line into the corresponding array (IPP or IPPS), then you call select() or poll() again to wait for the next output line, ... This way you pick up the data from both `ippfind` processes immediately as it appears, avoiding that any one of them gets blocked by its output not being collected. When both `ippfind` processes have ended (main process reads EOF) Go through the two arrays to build the final output list.

Could you implement one of these?

nidhijainpnp:
time taken after using select() to read ippfind output in CUPS array:
```
nidhi@nidhi-X510UNR:~$ time driverless
ipp://HP%20OfficeJet%20Pro%206960%20%5BC3B82D%5D%20(USB%201)._ipp._tcp.local/
ipps://HP%20OfficeJet%20Pro%206960%20%5BC3B82D%5D._ipps._tcp.local/
ipps://hello1._ipps._tcp.local/

real	0m2.042s
user	0m0.025s
sys	0m0.023s

nidhi@nidhi-X510UNR:~$ time driverless list
DEBUG: Started ippfind (PID 7252)
DEBUG: Started ippfind (PID 7253)
"driverless:ipp://HP%20OfficeJet%20Pro%206960%20%5BC3B82D%5D%20(USB%201)._ipp._tcp.local/" en "HP" "HP OfficeJet Pro 6960, fully driverless, cups-filters 1.27.5" "MFG:HP;MDL:OfficeJet Pro 6960;CMD:PCLM,PCL,PWGRaster,AppleRaster,JPEG,URF,PWG;"
"driverless:ipps://HP%20OfficeJet%20Pro%206960%20%5BC3B82D%5D._ipps._tcp.local/" en "HP" "HP OfficeJet Pro 6960, fully driverless, cups-filters 1.27.5" "MFG:HP;MDL:OfficeJet Pro 6960;CMD:PCLM,PCL,PWGRaster,AppleRaster,JPEG,URF,PWG;"
"driverless:ipps://hello1._ipps._tcp.local/" en "Example" "Example Printer, fully driverless, cups-filters 1.27.5" "MFG:Example;MDL:Printer;CMD:PWGRaster,AppleRaster,PWG,URF;"
DEBUG: PID 7252 (ippfind _ipps._tcp) exited with no errors.
DEBUG: PID 7253 (ippfind _ipp._tcp) exited with no errors.

real	0m4.002s
user	0m0.083s
sys	0m0.067s
```
time taken previously when one ippfind calls runs one after the other
(removed the output lines and only time is being shown here)
```
nidhi@nidhi-X510UNR:~$ time driverless
real	0m3.950s
user	0m0.025s
sys	0m0.035s

nidhi@nidhi-X510UNR:~$ time driverless list
real	0m4.896s
user	0m0.128s
sys	0m0.113s


```

tillkamppeter:
Thanks for your further work. I have merged now to also try to measure the execution times.
Unfortunately, I get a double free:
```
$ time ./driverless 
free(): double free detected in tcache 2
Aborted (core dumped)
```
Here is the `gdb -c core .libs/driverless` output:
```
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007f5ef09a0859 in __GI_abort () at abort.c:79
#2  0x00007f5ef0a0b44e in __libc_message (action=action@entry=do_abort, 
    fmt=fmt@entry=0x7f5ef0b35285 "%s\n") at ../sysdeps/posix/libc_fatal.c:155
#3  0x00007f5ef0a134dc in malloc_printerr (
    str=str@entry=0x7f5ef0b375d0 "free(): double free detected in tcache 2")
    at malloc.c:5347
#4  0x00007f5ef0a1514d in _int_free (av=0x7f5ef0b66b80 <main_arena>, 
    p=0x562dc46501d0, have_lock=0) at malloc.c:4201
#5  0x0000562dc31f20d0 in list_printers (mode=0, reg_type_no=<optimized out>, 
    isFax=644496684) at utils/driverless.c:786
#6  0x0000562dc31f08ff in main (argc=1, argv=0x7fff266a4298)
    at utils/driverless.c:976
(gdb) 
```
Could you check and fix that? Thanks.

tillkamppeter:
The crash seems to be cause by the duplicate removal of the arrays in lines 778 and 779. Removing the lines lets the crash go away for me, but I do not get any output of results, whereas the old version (what was in the master branch before merging this PR) shows the output correctly.

nidhijainpnp:
> The crash seems to be cause by the duplicate removal of the arrays in lines 778 and 779. Removing the lines lets the crash go away for me, but I do not get any output of results, whereas the old version (what was in the master branch before merging this PR) shows the output correctly.
On my system, I am getting the correct output results after removing the lines too..I am still figuring out why you don't get any output results and would let you know as soon as I found the glitch.

nidhijainpnp:
Hi,
could you please try to change the 
```
timeout.tv_sec  = 2;
 timeout.tv_usec = 0;
```
values to something higher like 5 or so:
```
timeout.tv_sec  = 5;
 timeout.tv_usec = 0;
```

tillkamppeter:
Yes, the 5-second timeout works, but I have found a flaw. If the timeput is reached you simply stop the output collection process. The better way is to do the following:
```
    timeout.tv_sec  = 0;
    timeout.tv_usec = 200000;

    if (select(nfds, &input, NULL, NULL, &timeout) < 0)
      return (-1);
    if (!FD_ISSET(read_ipp_fd, &input) && !FD_ISSET(read_ipps_fd, &input))
    {
      continue;
    } 
```
This shortens the timeout to 200 msec but if the timeout is reached, the select() call is simply repeated. This way we cope with any speed behavior of `ippfind`.
Now theoretically the two `ippfind` calls shoyuld not block each other, but my measurements give the impression that they are still running one after the other. Could you check?

utils/driverless.c
```diff
@@ -59,6 +59,142 @@ convert_to_port(char *a)
...
+      
+    }
+
+    if(FD_ISSET(read_ipp_fd, &input))
+    {
+      dup2(read_ipp_fd, 0);
+      fp = cupsFileStdin();
+      if ((bytes = cupsFileGetLine(fp, buffer, sizeof(buffer))) > 0) 
+      {
+        ippfind_output = (char *)malloc(MAX_OUTPUT_LEN*(sizeof(char)));
```
zdohnal, line 159: @nidhijainpnp please free the memory you allocate.

utils/driverless.c
```diff
@@ -59,6 +59,142 @@ convert_to_port(char *a)
...
+    
+    if(FD_ISSET(read_ipps_fd, &input))
+    {
+      dup2(read_ipps_fd, 0);
+
+      fp = cupsFileStdin();
+
+     if ((bytes = cupsFileGetLine(fp, buffer, sizeof(buffer))) > 0) 
+      {
+        ippfind_output = (char *)malloc(MAX_OUTPUT_LEN*(sizeof(char)));
```
zdohnal, line 118: @nidhijainpnp please free the memory you allocate.

utils/driverless.c
nidhijainpnp: Hi,
I have to remove those two lines 
```
free(ippfind_output);
ippfind_output = NULL;
```
because it gives an error on testing:
```
nidhi@nidhi-X510UNR:~$ driverless 
Segmentation fault (core dumped)
```
Could you please confirm me if you are getting the same error on adding those two lines (freeing up the memory by `ippfind_output`)?

zdohnal: Hi @nidhijainpnp ,

hmm, I don't get any segfault if I add the lines - with and without any driverless device on the local network. It would be good to see a segfault backtrace, then we can be able to figure out where the error is. 

The interesting is ```valgrind``` is not capable to find any leaks - my guess it is due multi-process - but if you create a minimal test case:

```
#include <stdlib.h>
#include <stdio.h>
#include <cups/array.h>
#include <string.h>


int main()
{

  char *i = (char*)malloc(10*sizeof(char));
  memset(i, 0, 10);

  cups_array_t *a = cupsArrayNew(NULL, NULL);

  cupsArrayAdd(a, i); 

  cupsArrayDelete(a);

  return 0;
}
```

, compile it with ```gcc -g -O0 -std=gnu11 -pedantic -Werror -Wall -o <output_binary> <source_file> -lcups``` and run ```valgrind --leak-check=full ./<output_binary>```, you will see a report for a leaking element.

If you look into ```cups/array.c``` of CUPS project, you see:
```
/*
 * 'cupsArrayDelete()' - Free all memory used by the array.
 *
 * The caller is responsible for freeing the memory used by the
 * elements themselves.
```

And ```service_uri``` is allocated on the heap too, so it should be freed.

zdohnal: @nidhijainpnp you can pull the newest pull request which puts those lines back instead of manual copying.

zdohnal: And I see you call:

```
nidhi@nidhi-X510UNR:~$ driverless 
```

Do you compile cups-filters master branch and install it via ```sudo make install``` or install the newest changes by other way?
Because IIUC this way you call the system ```driverless``` binary:

```
$ which driverless
/usr/bin/driverless
```

instead of ```driverless``` binary you built within project repository:

```
$ pwd
/home/zdohnal/repo_upstream/cups-filters
$ ls -lah driverless
-rwxrwxr-x. 1 zdohnal zdohnal 7.7K Sep  7 07:29 driverless
$ which ./driverless
./driverless
```

So the older ```driverless``` binary from system is called, unless you install the newest ```driverless``` from the project before.

PR #306, opened 2020-09-07
Free the used memory

zdohnal:
Memory cleanup after the last merge.

P.S. It can be an advantage if students are encouraged to use ```valgrind --leak-check=full``` when testing their PRs - it can find memory leaks and double free errors.

### DIFF
--- a/utils/driverless.c
+++ b/utils/driverless.c
@@ -490,9 +490,11 @@ listPrintersInArray(int reg_type_no, int mode, int isFax, char* ippfind_output )
 	} 
 
       read_error:
+        free(service_uri);
         return;
       }
 
+  free(service_uri);
   return;
 }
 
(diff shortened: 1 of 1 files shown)


utils/driverless.c
```diff
@@ -125,6 +125,9 @@ copy_output(int read_ipp_fd,		/* I - IPP file descriptor */
           continue;
         snprintf(ippfind_output, MAX_OUTPUT_LEN, "%s", ptr);
         cupsArrayAdd(service_uri_list_ipps,ippfind_output);
+
+        free(ippfind_output);
+        ippfind_output = NULL;
```
tillkamppeter, line 130: AFAIR cupsArrayAdd() only adds a pointer to the array and does not copy/strdup the string. So if you free the string immediately afterwards as you do here, you end up with an array of invalid pointers.
The freeing has be done when the after using the array, right before deleting it. Or you define the array with a freeing function, then deleting it will also free the content.

zdohnal: @tillkamppeter Yes, you're right - my test sample was incorrect (I wasn't accessing the element from the array afterwards...), which led me to think the element needs to be freed before other iteration to prevent losing an allocated pointer.
@nidhijainpnp sorry for the noise about array elements - my proposed fix is incorrect, I'll define arrays with ```cupsArrayNew3()``` which allows to send freeing function as a parameter.

Though freeing ```service_uri``` string is only thing correct in the PR...

I'll be back in a while.

zdohnal:
@tillkamppeter the code is updated to use ```cupsArrayNew3()``` which allows definig freeing function. Thank you for the review!

PR #311, opened 2020-10-12
cups-browsed.c: uuid was used after its pointer was freed by ippDelete()

zdohnal:
Hi,

valgrind [results](https://bugzilla.redhat.com/attachment.cgi?id=1717802) in Fedora [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1881365) revealed that I didn't understand how ```ippFindAttribute()``` works when I issued a PR which clears ```response``` variable in ```get_printer_uuid()```. The ```ippFindAttribute()``` returns a pointer to a member of ```response``` structure and this pointer is freed when ```response``` is freed with ```ippDelete()```.

The fix now uses ```strdup()``` for the pointer and this copy is returned as return value.

### DIFF
--- a/utils/cups-browsed.c
+++ b/utils/cups-browsed.c
@@ -3614,7 +3614,7 @@ new_local_printer (const char *device_uri,
 {
   local_printer_t *printer = g_malloc (sizeof (local_printer_t));
   printer->device_uri = strdup (device_uri);
-  printer->uuid = (uuid ? strdup (uuid) : NULL);
+  printer->uuid = uuid;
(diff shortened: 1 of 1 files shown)


tillkamppeter:
Thank you very much for this fix.

tillkamppeter:
@zdohnal, the PR you ment where you introduced the problem was #281? Am I right?

zdohnal:
@tillkamppeter Yes, that's the commit, I'm sorry for inconvenience.

Issue #326, opened 2020-12-03
Change handling of "Gray" to match CUPS driverless behavior


IslandCow:
When handling for "Gray" was added to cupsRasterParseIPPOptions, it was mapped to 3 or CUPS_CSPACE_K.  However, in CUPS PPD generation, that was only used for bi-level.  Gray (what was generated for monochrome) should be mapped to CUPS_CSPACE_SW or 18.  For convenience, the reference from CUPS is included below.

https://github.com/apple/cups/blob/9240cd7/cups/ppd-cache.c#L3767

tillkamppeter:
Thanks for the bug report and pull request. I have fixed this now for more, hopefully all cases where the PPD generator and cupsRasterParseIPPOptions() mismatch, commit 1dc90ce0f.

PR #327, opened 2020-12-03
Remap Gray to CUPS_CSPACE_SW instead of CUPS_CSPACE_K.

IslandCow:
Fix for #326


tillkamppeter:
Applied more complete fix, commit 1dc90ce.
But thanks for the pull request anyway.

Issue #314, opened 2020-10-27
cups-browsed gets SIGSEGV when generating PPD file


zdohnal:
Hi,

we have an issue in [Fedora](https://bugzilla.redhat.com/show_bug.cgi?id=1881365) about ```cups-browsed``` getting SIGSEGV when generating PPD file.

```cups-browsed``` is set to ```BrowsePoll``` to CentOS 7 server, where a print queue with following [ppd](https://bugzilla.redhat.com/attachment.cgi?id=1723493) is installed. CUPS-1.6.3 from CentOS is not capable to return ```media-col-database``` via response, so ```Get-Printer-Attributes``` operation passes only with ```All``` keyword for requested attributes.

The attached [gdb log](https://bugzilla.redhat.com/attachment.cgi?id=1723213) shows ```ppdname``` contains ```A4``` because ```media-col-default``` contains non-borderless value, but other attributes are missing (```media-col-database```) or are set as borderless (```media-size-supported```, ```media-supported```).

Then the SIGSEGV comes in this code:
```
2224 
2225     if (all_borderless) {
2226       suffix = strcasestr(ppdname, ".Borderless");
2227       *suffix = '\0';
2228     }
```

```all_borderless``` is still ```1```, because all medias in sizes are ```Borderless```, but ```ppdname``` is ```A4```, so it segfaults on ```*suffix = '\0';``` because suffix is NULL.

The quick fix would be to check whether ```suffix``` is NULL, but it feels like narrow fix - what do you think about how the code should behave in situations when ```media-col-default``` tells non-borderless media and other medias are borderless?

tillkamppeter:
Simple addition of NULL checks is OK here, as we only want to remove unneeded ".Borderless" suffixes.
Fixed in commit 240ffb901d in master. Merge to 1.x will come later.

PR #328, opened 2020-12-07
Cherry-pick the fix for #314 and push it to 1.x

zdohnal:
SSIA


tillkamppeter:
Thank you very much. I have applied it by directly using `git cherry-pick -x 240ffb9` though. Ended up as commit e33e464d3 in 1.x.

PR #333, opened 2020-12-29
pdftopdf : Fix no final_content_type environment variable found bug

pranshukharkwal:
Upon running pdftopdf filter, 
`DEBUG: pdftopdf: No FINAL_CONTENT_TYPE environment variable, could not determine whether to log pages or not, so turned off page logging.`

This Debug statement was being printed. This was because in pdftopdf.c , NULL was being sent to the parameters argument instead of the final output type. 
So the fix is to send `OUTPUT_FORMAT_PDF` in the argument, instead of NULL.

### DIFF
--- a/filter/pdftopdf.c
+++ b/filter/pdftopdf.c
@@ -52,9 +52,8 @@ main(int  argc,				/* I - Number of command-line args */
 #else
   signal(SIGTERM, cancel_job);
 #endif /* HAVE_SIGSET */
-  filter_out_format_t outformat = OUTPUT_FORMAT_PDF;
-  ret = filterCUPSWrapper(argc, argv, pdftopdf, &outformat, &JobCanceled);
-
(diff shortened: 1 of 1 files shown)


tillkamppeter:
Did you test this? This is not correct. `final_content_type` is a string variable, which is expected to hold the content of the `FINAL_CONTENT_TYPE` environment variable, whereas `OUTPUT_FORMAT_PDF` is a value of an `enum` type. Note that the parameters pointer can point to completely different data types for each filter function, it is for filter-specific data. You need to put a pointer to the value of the `FINAL_CONTENT_TYPE` environment variable as the `parameters` argument in the call in `filter/pdftopdf.c`.

pranshukharkwal:
@tillkamppeter I have made the required changes. Please check now

tillkamppeter:
That is correct now. Thank you very much. I have merged it.

PR #336, opened 2021-01-07
adding files in gitignore that are generated by autogen.sh configure …

surajkulriya:
…and make


tillkamppeter:
It is merged now, thank you very much.

PR #337, opened 2021-01-09
cups-browsed.service: Adapt to changes in CUPS

renodr:
Upstream CUPS seems to install org.cups.cupsd.service now instead of cups.service.

### DIFF
--- a/utils/cups-browsed.service
+++ b/utils/cups-browsed.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=Make remote CUPS printers available locally
-Requires=cups.service
-After=cups.service avahi-daemon.service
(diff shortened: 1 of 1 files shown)


tillkamppeter:
Which version of CUPS are you using and from where do you get it?
Currently maintained upstream of CUPS is [here ate OpenPrinting](https://github.com/OpenPrinting/cups) as Apple seems to have abandoned CUPS development, see my [monthly news posts](https://openprinting.github.io/news/), especially [September 2020](https://openprinting.github.io/OpenPrinting-News-September-2020/) and following.
On OpenPrinting we have renamed the CUPS service to `cups.service` following the Linux distributions and so cups-browsed as it is now is correct.

PR #339, opened 2021-01-22
updated final_content_type in filters imagetopdf, imagetoraster and p…

surajkulriya:
updated final_content_type in filters imagetopdf, imagetoraster and pclmtoraster

### DIFF
--- a/filter/imagetopdf.c
+++ b/filter/imagetopdf.c
@@ -64,7 +64,8 @@ main(int  argc,				/* I - Number of command-line args */
   * Fire up the imagetopdf() filter function
   */
 
-  ret = filterCUPSWrapper(argc, argv, imagetopdf, NULL, &JobCanceled);
+  char *t = getenv("FINAL_CONTENT_TYPE");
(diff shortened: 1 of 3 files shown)


tillkamppeter:
Thank you for the pull request, but I will not merge it as it has no effect at all.

You are feeding the value of the `FINAL_CONTENT_TYPE` environment variable into the filters via the parameters pointer, simply pointing to the string containing the value of said variable. Probably you have taken this idea from the `pdftopdf()` filter function.

But note that the parameters have no standard meaning, they have individual purposes varying from filter function to filter function. If they had the same meaning for all filter functions I had accommodated them as fields of the `filter_data` record. Have a look into the source code of the filter functions. The `parameters` argument of the main function is used for very different purposes (documented in `cupdfilters/filter.h`). Only the `pdftopdf()` filter function expects the value of the `FINAL_CONTENT_TYPE` environment variable, in many other filter functions (including all considered by you in this PR) the `parameters` argument is not used at all. 

If you think that any of the other filter functions except `pdftopdf()` should behave different depending on the value of the `FINAL_CONTENT_TYPE` environment variable you need to implement this in the filter function first.

PR #356, opened 2021-04-01
margins updated texttopdf

surajkulriya:
So the issue is related to texttopdf filter. When we run it without a ppd file, it creates a pdf file with many no of pages, all empty. Below changes seems to work. 
@tillkamppeter  can you please review it. 
Thanks

### DIFF
--- a/cupsfilters/raster.c
+++ b/cupsfilters/raster.c
@@ -594,6 +594,9 @@ cupsRasterParseIPPOptions(cups_page_header2_t *h, /* I - Raster header */
   else if (set_defaults)
   {
     /* TODO: Automatic A4/Letter, like in scheduler/conf.c in CUPS. */
+    h->cupsPageSize[0] = 612.0f;
+    h->cupsPageSize[1] = 792.0f;
+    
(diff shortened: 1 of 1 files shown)


tillkamppeter:
Could you tell me which reported issue gets fixed by this?

surajkulriya:
Actually the thing is that issue is not reported. I was just checking whether texttopdf filter works fine without ppd file or not. It didn't.
Then I looked deeper into this and found out default values of `PageLeft`, `PageRight`, `PageBottom`, `PageTop`, `PageWidth` and `PageLength` were all zeroes in case PPD file=NULL in `pagecommon.c` file (Line no 617).  So I just changed the default value of these variables.
To reproduce the issue I faced, please first make sure that environment variable PPD is not defined. then , run command-
`/usr/lib/cups/filter/texttopdf 1 user title 1 "" input > output.pdf`.
You will see that output.pdf contains only empty pages.
Given PR is fix for this.

tillkamppeter:
@surajkulriya thank you very much for finding the bug and fixing it. I have merged your changes now.

PR #359, opened 2021-04-15
cups-browsed: Save default options

bmasonrh:
cups-browsed doesn't currently save printer options with the suffix "-default".  For example running:

`lpadmin -p my-queue -o lpi-default=18`

will change the defaults, but then the default doesn't persist when cups-browsed is restarted.


bmasonrh:
I've verified that the patch fixes the issue in cups-filters 1.20.0 in RHEL 8.  The affected code is the same.

tillkamppeter:
solved a little bit differently with cdd61132e1.

Thank you for your pull request.

Issue #347, opened 2021-02-27
ppi option is an int, but image2raster and image2pdf treat it as a string


kfazz:
for printing images with asymmetric dpi image2raster and image2pdf suppport the form ppi=300x600 to set xppi and ypi , however
ppi is defined by cups as an int, so the option will be sanitized prior to being passed to the filters. 
Therefore the image* filters should add an additional yppi option to set the yppi in addition to the xppi

current code:
```
  if ((val = cupsGetOption("ppi", num_options, options)) != NULL)
  {
    if (sscanf(val, "%dx%d", &xppi, &yppi) < 2)
      yppi = xppi;
    zoom = 0.0;
  }
```
a potential fix might be like this:
```
  if ((val = cupsGetOption("ppi", num_options, options)) != NULL)
  {
     xppi = atoi(val);
    if ((val = cupsGetOption("yppi", num_options, options)) != NULL)
      {
         yppi = atoi(val);
      } else {
         yppi = xppi;
      }
    zoom = 0.0;
  }
```

reported to cups here:
 https://github.com/OpenPrinting/cups/issues/115
original ubuntu report here:
https://bugs.launchpad.net/cups/+bug/1917074

pranshukharkwal:
@kfazz Could you share the png file you are working with, so that it will be easier to replicate the exact issue

tillkamppeter:
Fixed by pull request #361, commit 5bb90d5cd.

PR #361, opened 2021-04-20
remove support for asymmetric ppi in imagesto___ filters and use ppi=…

pranshukharkwal:
…XXX as integer as it is in cups

Solves #347


tillkamppeter:
Thank you very much, it is merged now.

PR #363, opened 2021-04-25
Pdftoraster: free colour data after processing of each page

pranjanpr:
This PR is related to the bug :- https://bugs.launchpad.net/ubuntu/+source/cups-filters/+bug/1920190 
RAM(4 GB) improves from 32% usage to 7% for pdftoraster filter.

### DIFF
--- a/filter/pdftoraster.cxx
+++ b/filter/pdftoraster.cxx
@@ -1231,8 +1231,8 @@ static void writePageImage(cups_raster_t *raster, poppler::document *doc,
         bp += rowsize;
       }
     }
-    free(colordata);
   }
+  free(colordata);
   if (allocLineBuf) delete[] lineBuf;
 }
 
--- a/filter/pdftoraster.cxx
+++ b/filter/pdftoraster.cxx
@@ -1231,6 +1231,7 @@ static void writePageImage(cups_raster_t *raster, poppler::document *doc,
         bp += rowsize;
       }
     }
+    free(colordata);
   }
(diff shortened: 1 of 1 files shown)


manulari:
Looking at the function where this change was made (writePageImage) there seem to be quite a few more mallocs that are never followed by a free in the case where the color space is some variant of gray.
Starting here:
https://github.com/pranjanpr/cups-filters/blob/7a4198f0c3ee3e645b1cdd216995cab02945e3f9/filter/pdftoraster.cxx#L1164


And what's up with this weird mix of mostly malloc and then a single use of new and an overridden allocator for new?:
https://github.com/pranjanpr/cups-filters/blob/7a4198f0c3ee3e645b1cdd216995cab02945e3f9/filter/pdftoraster.cxx#L1202
https://github.com/pranjanpr/cups-filters/blob/7a4198f0c3ee3e645b1cdd216995cab02945e3f9/filter/pdftoraster.cxx#L1635

adhiti33:
@manulari Quick question: Regarding pdftoraster, how about timetaken to process and out raster filesize in worstcase ?

fitzsim:
Thank you for this fix.  It made my `Brother HL-L2390DW` able to print PDF files with images in them.  The fix has not made it into `Debian 12` yet (`cups-filters-core-drivers 1.28.7-1+deb11u3`), so I compiled it in myself.  Without this patch, `pdftoraster` hangs taking 100% CPU (never checked memory before killing it), and the print job never completes.  I hope the `Debian` packages incorporate this in their next update.

PR #364, opened 2021-04-26
utils/cups-browsed.c: Lock before all 'examine_discovered_printer_rec

zdohnal:
…ord()'

The current code doesn't use rwlock in found_cups_printer() and
cupsdUpdateLDAPBrowse(), which causes deadlock when discovering queues
via BrowsePoll or LDAP.
It happens because the code in 'examine_discovered_printer_record()'
expects the rwlock to be locked already, so it unlocks the lock, calls
'get_local_queue_name()' and locks again.

The commit contains change '&(lock)' -> '&lock' too, it represents the
same thing, so we should use the same code for it.

### DIFF
--- a/utils/cups-browsed.c
+++ b/utils/cups-browsed.c
@@ -4833,9 +4833,11 @@ cupsdUpdateLDAPBrowse(void)
     debug_printf("LDAP: Remote host: %s; Port: %d; Remote queue name: %s; Service Name: %s\n",
 		 host, port, strchr(local_resource, '/') + 1, service_name);
 
+    pthread_rwlock_wrlock(&lock);
     examine_discovered_printer_record(host, NULL, port, local_resource,
 				      service_name, location, info, "", "",
 				      "", 0, NULL);
+    pthread_rwlock_unlock(&lock);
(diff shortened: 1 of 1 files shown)


tillkamppeter:
Lools like that the original multi-threading code was only tested under "modern" conditions (all CUPS in local network >= 1.6.x, DNS-SD-only).

Thank you very much for fixing the legacy cases.

zdohnal:
@tillkamppeter since there is still no other simple (you can theoretically reflect mDNS or create a container within your LAN which is basically a copy of remote server, but it is more difficult than just one-liner on cups-browsed) and full-fledged (you can set ServerName in /etc/cups/client.conf, but you cannot modify the queues unless you have permissions on the server) solution to get a print queue from a different subnet than what cups-browsed provides, we're stuck with it until we implement something new in CUPS - so it doesn't sound as a legacy case. Most users which aggregate their printers at servers in one LAN and have their PCs in other LAN use BrowsePoll.

I'm not sure about LDAP, I didn't test that one - but it looked it can suffer the same problem.

Issue #354, opened 2021-03-22
gstoraster: Ghostscript runs to long while converting djvu files to printer`s format


valia0906:
https://bugs.launchpad.net/ubuntu/+source/cups-filters/+bug/1920730

tillkamppeter:
@valia0906, could you attach a sample djvu file and also instructions how you print the djvu file, so that we can reproduce the problem and perhaps improve the patch somewhat? This is especially important as we want to carry on the CUPS filters when CUPS will not use PPD files any more.

If you are the poster of the Ubuntu bug or if you have at least tried out the patch posted there and found out that it solves the problem. could you try the following:

As the Ghostscript call the patch is modifying does not more than counting the pages (there is another one for the actual rasterization) could you also try to modify the patch by not calling cupsRasterInterpretPPD() for extracting the default page size from the PPD and instead, simply insert arbitrary height and width, like "-dDEVICEWIDTHPOINTS=612 -dDEVICEHEIGHTPOINTS=792"? This would allow you to remove the "if (ppd)" and so your fix would also work for using the filter without PPD.

Could you try this out? Do you still get fast execution and correct page counts (= filter outputs all pages)?

@surajkulriya, if you are able to reproduce the original problem (having an input file which makes the original filter taking to long but the patch solving this) and did not simply grab the patch, could you also try out what I am suggesting above? This would also be a first step for your GSoC project.

tillkamppeter:
@surajkulriya I got a test file (see Ubuntu bug report) and tried it as I described there. I found that the best solution is to simply add `-dDEVICEWIDTHPOINTS=1 -dDEVICEHEIGHTPOINTS=1` to the command line:
```
gs -q -dNOPAUSE -dBATCH -sDEVICE=bbox -dDEVICEWIDTHPOINTS=1 -dDEVICEHEIGHTPOINTS=1 sample2.pdf 2>&1 | grep -c HiResBoundingBox
```
So no `if (ppd)` and no call of `cupsRasterInterpretPPD()` is needed, and the fix also applies to PPD-less use of the filter.

surajkulriya:
@tillkamppeter  sir, 
Thanks for the update, I have made changes in the latest commit.

tillkamppeter:
Fixed by pull request #371.

Thank you very much.

PR #371, opened 2021-05-13
made changes to convert djvu files to printer's format faster

surajkulriya:
fixed issue #354 
@tillkamppeter sir, can you please review this
many thanks

### DIFF
--- a/cupsfilters/ghostscript.c
+++ b/cupsfilters/ghostscript.c
@@ -853,15 +853,9 @@ ghostscript(int inputfd,         /* I - File descriptor input stream */
     size_t bytes;
     // Ghostscript runs too long while converting djvu files to Xerox`s 3210 format
     // Using -dDEVICEWIDTHPOINTS -dDEVICEHEIGHTPOINTS params solves the problem
-    if (ppd) {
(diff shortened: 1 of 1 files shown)


tillkamppeter:
@surajkulriya Thank you very much. I have merged your pull request now, but with corrected commit message.

Note that the command line you have worked on does not convert DjVu to PDF. evince calls some utility to convert the DjVu file to PDF when printing. Probably these PDFs have something odd which makes the bbox device of Ghostscript taking so long. The bbox device is abused here to count pages as it does not output megabytes of raster data, but for its original purpose of determining the bounding box of  each page it seems that it is fast if the pages have a well-defined bounding box, which is the case if you supply `-dDEVICEWIDTHPOINTS=1 -dDEVICEHEIGHTPOINTS=1` on the command line, otherwise, in a worst case if the PDF does not contain bounding boxes, it has to render the graphics on each page (which is probably the case with these PDFs).

Smaller numbers for the fake page size seem to accelerate even more, probably even with a fixed page size the graphics falling into that page is somehow investigated and if the pages are a tiny area at the lower left corner there is usually no graphics at all, but for only counting the pages it is good enough.

PR #377, opened 2021-06-17
cups-browsed.c: Introduce NotifLeaseDuration directive

zdohnal:
The PR implements two things:

1) NotifLeaseDuration directive for cups-browsed.conf - it will make
   lease duration for notifications configurable by users. IMO it is not
   useful for regular users, but it is helpful during sanity testing
   (for verifying that we actually renew the subscription when time
   comes). The current hardcoded 1 day is unusuable for that :( .
   I implemented the lowest threshold to 300s to prevent a possible DoS.
2) Subscription renewal is set to happen in the middle of NotifLeaseDuration,
   not one minute before lease expiration. This was a problem on busy servers,
   where cups-browsed was busy and wasn't able to renew the subscription
   before cupsd removed it. Then if some jobs had come before the subscription
   was created again, the queue got disabled. The proposed approach is based
   on behavior of DHCP.

### DIFF
--- a/utils/cups-browsed.c
+++ b/utils/cups-browsed.c
@@ -142,7 +142,6 @@ static int  ldap_rebind_proc(LDAP *RebindLDAPHandle,
 #define TIMEOUT_REMOVE      -1
 #define TIMEOUT_CHECK_LIST   2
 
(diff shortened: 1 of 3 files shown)


zdohnal:
Sorry for the noise, I'll return with the PR tomorrow.

PR #379, opened 2021-06-21
make use of ipp attributes in rastertopclm and rastertopdf filters if PPD file is not present

surajkulriya:


tillkamppeter:
Thank you very much, I have merged your work now.

PR #381, opened 2021-06-24
Converted pdftoraster filter to filter function.

pranjanpr:


tillkamppeter:
Thank you very much for the pull request, but I have found some issues:
- It seems that to add new files to the repo you did a wholesale `git add *` instead of adding the actual files individually, as you have added a lot of stray files,m esp[ecially the ones generated by a test run (`make check`?). Make sure that only human-generated files make it into the repository.
- Make sure that there is no `fprintf(stderr, ...)` in the filter function and any function called by it. Let all logging go through the log function.
- Make sure that the last line ends with a newline and that you do not add trailing white space to any of the lines.
- Please rename the CUPS filter stub in the `filter/` directory to `pdftoraster.c` as it is not C++ any more.

Best is probably to create a new, clean pull request on current master state of cups-filters. I will probably not commit to cups-filters in the next 24 hours (I am onm ps-printer-app currently). This should facilitate your pull request.

tillkamppeter:
Assuming you did not actively modify the test suite, you modified the following files:
- `Makefile.am`
- `cupsfilters/filter.h`
- `filter/pdftoraster.cxx`

and you added the following file:
- `cupsfilters/pdftoraster.cxx`

All the other changes/additions, seem to come from a `make check` run.

So in your new pull request on a fresh branch ONLY do
```
git add cupsfilters/pdftoraster.cxx
```
to add your new file and do not use wildcards here.

PR #382, opened 2021-06-25
pdftoraster filter converted to filter function

pranjanpr:
I will look into changing the extension of `filter/pdftoraster.cxx` to `filter/pdftoraster.c`. Changing it initiates `make` issues so currently it is in the form of `.cxx`. Will resolve soon.


tillkamppeter:
Fixed the .c extension problem.

Important is (I assume a part of these things you already did on your attempts):
- Rename all occurences of `pdftoraster.cxx` to `pdftoraster.c` in Makefile.am
- The CUPS filter stub does not need all the libraries, include files and C/C++ flags which the filter itself needs, so in `Makefile.am` they all can get removed in the part for building the `filter/pdftoraster.c`
- Simply running `make` again after editing `Makefile.am` does not trigger all new autoconf/automake/libtool runs needed. You need to rung `make distclean` then `./configure` and after that it works.

The new commit is: b0584bddf

pranjanpr:
Thanks for elaborating sir. Will keep these points in mind while modifying other filters.

tillkamppeter:
The move of the code into libcupsfilters caused some compiler warnings (not your changes, probably simply the move into the library) which I have fixed now: d0fee95b6

tillkamppeter:
Also note for log messages that if they are handled by a log function that the log function adds the newline at the end of the line if needed, you do not need to specify it. So on converting `fprintf(stderr, ...)` to log function calls, remove the `'\n'` in the end.

Fixed in 9c1fb4b09

tillkamppeter:
Updated also the NEWS file: a4834e0e4

tillkamppeter:
THenk you very much for this contribution.

Well done! You have completed your first transition of a classic CUPS filter into a filter function.

About the formatting fixes which I have done in my last few commits here note that they are not necessarily all your fault, many I have probably only discovered now when I have looked through the code and your changes.

PR #388, opened 2021-07-04
Rastertopwg filter converted to filter function.

pranjanpr:


cupsfilters/rastertopwg.c
```diff
@@ -0,0 +1,558 @@
...
+		   "rastertopwg: Unable to open output data stream.");
+    }
+
+    cupsFileClose(inputfp);
+
+    return (1);
+  }
+
+
+  if ((final_content_type = getenv("FINAL_CONTENT_TYPE")) == NULL)
```
tillkamppeter, line 99: Do not use `getenv()` in a filter function, instead, read the variable in the CUPS filter stub (`filter/rastertopwg.c`) and transfer the requested output format as a parameter. See the `ghostscript()` and `pdftopdf()` filter functions and the CUPS filter stubs `gstoraster` and `pdftopdf`. You can also use the type `filter_out_format_t` for the parameter here.

cupsfilters/rastertopwg.c
```diff
@@ -0,0 +1,558 @@
...
+    final_content_type = "image/pwg-raster";
+
+  inras  = cupsRasterOpen(inputfd, CUPS_RASTER_READ);
+  outras = cupsRasterOpen(1, !strcmp(final_content_type, "image/pwg-raster") ? CUPS_RASTER_WRITE_PWG : CUPS_RASTER_WRITE_APPLE);
+
+    if (data->ppdfile == NULL && data->ppd == NULL)
+  {
+    char *p = getenv("PPD");
```
tillkamppeter, line 107: The PPD you get via the filter data (`data->ppd`) here you also do not need to use `getenv()` and also should not use it.

tillkamppeter:
Did you also test the code to see whether it is stable and produces the correct output? You can visualise CUPS/PWG/Apple Raster with `rasterview` from [here](https://www.msweet.org/rasterview/).

tillkamppeter:
@pranjanpr Looks all OK now, thanks. Please test it.

tillkamppeter:
@pranjanpr did you also check whether there are no memory leaks? This means that if dynamic memory is allocated that it gets freed before the filter function returns. This is especially important when this filter function is used in a permanently running Printer Application.

tillkamppeter:
Your input file has to be CUPS Raster, not PWG or Apple Raster. rasterview displays all three flavors. So if you display your input file with rasterview, click the button at the lower right corner of the window to fold out the metadata panel. At the top it should tell them which of the three formats your input file is.

Then run `imagetoraster` with appropriate command line options to make it output CUPS Raster. Perhaps you use a PPD file (of HPLIP, SpliX, Gutenprint, ...) with a `*cupsFilter` line which tells that the requested input format is `application/vnd.cups-raster`. Run your filter with that PPD file (set `PPD` env variable to it).

Setting the env variable `FINAL_CONTENT_TYPE=application/vnd.cups-raster` probably also works. Also set the option `Pagesize=A4` to have a defined output page size.

pranjanpr:
@tillkamppeter 
Sir I have tested the filter function. It is running smoothly. Got the desired output from `rastertopwg`.

pranjanpr:
Also resolved memory leaks which valgrind was pointing out.

tillkamppeter:
@pranjanpr Thank you very much.

I have fixed some compiler warnings (c79527bde), please check for warnings on further conversions.

PR #393, opened 2021-07-07
silenced compiler warning in pdftops filter

surajkulriya:


tillkamppeter:
Thank you very much. It works correctly now.

PR #407, opened 2021-08-14
Bannertopdf converted into filter function.

pranjanpr:
Segfault issue resolved.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -50,7 +50,6 @@ pkgbackend_PROGRAMS = parallel serial beh implicitclass
 parallel_SOURCES = \
 	backend/parallel.c
 parallel_LDADD = \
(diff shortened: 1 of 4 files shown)


Makefile.am
```diff
@@ -50,6 +50,7 @@ pkgbackend_PROGRAMS = parallel serial beh implicitclass
 parallel_SOURCES = \
 	backend/parallel.c
 parallel_LDADD = \
+	libfontembed.la \
```
tillkamppeter, line 53: Make sure that this line does not get added back in, it is wrong and therefore I have removed it in commit 8e91b0b572 and did the correct thing to trigger the build of libfontembed via commit d33583e01,

pranjanpr: Updated PR

PR #408, opened 2021-08-18
print-render support added in ghostscript and imagetoraster function

surajkulriya:
@tillkamppeter  sir please review it.
Thank you very much

### DIFF
--- a/cupsfilters/ghostscript.c
+++ b/cupsfilters/ghostscript.c
@@ -810,7 +810,6 @@ ghostscript(int inputfd,         /* I - File descriptor input stream */
   void          *ld = data->logdata;
   filter_iscanceledfunc_t iscanceled = data->iscanceledfunc;
   void          *icd = data->iscanceleddata;
-  const char *val;
 
 
   /* Note: With the OUTPUT_FORMAT_APPLE_RASTER selection the output is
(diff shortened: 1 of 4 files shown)


tillkamppeter:
Principally the code is correct.
But it looks like you have pasted one big piece of code identically into two filter functions. Could you check whether you can define a function with this code and call it in the two filter functions?

surajkulriya:
updated sir

tillkamppeter:
Thank you very much. It is merged now.

For consistency, we need this also in pdftoraster() and mupdftoraster()

surajkulriya:
Sir,
I think its already been added in pdftoraster in my PR while making it PPD independent.
I will add this support in mupdftoraster filter function too.
Thank you very much

tillkamppeter:
Are you using your new function in pdftoraster()? Or is pdftoraster duplicating that code? Or is the function not applicable there for some reason?

surajkulriya:
sir actually the code is slightly different there in that case. In pdftoraster filter function we are storing information in pdftoraster_doc_t structure that is different from cups_page_header2_t structure. So I think the function defined will not work for pdftoraster filter function.
Thanks

tillkamppeter:
In `pdftoraster()` you have
```
    cupsRasterParseIPPOptions(&(doc->header),data,doc->pwgraster,1);
    if (strcasecmp(doc->header.cupsRenderingIntent, "Perceptual") == 0) {
      doc->colour_profile.renderingIntent = INTENT_PERCEPTUAL;
    } else if (strcasecmp(doc->header.cupsRenderingIntent, "Relative") == 0) {
      doc->colour_profile.renderingIntent = INTENT_RELATIVE_COLORIMETRIC;
    } else if (strcasecmp(doc->header.cupsRenderingIntent, "Saturation") == 0) {
      doc->colour_profile.renderingIntent = INTENT_SATURATION;
    } else if (strcasecmp(doc->header.cupsRenderingIntent, "Absolute") == 0) {
      doc->colour_profile.renderingIntent = INTENT_ABSOLUTE_COLORIMETRIC;
    }
```
and afterwards you evaluate the option/attribute and change the rendering intent setting in `doc` appropriately.
Would it not also work if youdo the following instead:
```
    cupsRasterParseIPPOptions(&(doc->header), data ,doc->pwgraster, 1);
    getPrintRenderIntent(data, &(doc->header));
    if (strcasecmp(doc->header.cupsRenderingIntent, "Perceptual") == 0) {
      doc->colour_profile.renderingIntent = INTENT_PERCEPTUAL;
    } else if (strcasecmp(doc->header.cupsRenderingIntent, "Relative") == 0) {
      doc->colour_profile.renderingIntent = INTENT_RELATIVE_COLORIMETRIC;
    } else if (strcasecmp(doc->header.cupsRenderingIntent, "Saturation") == 0) {
      doc->colour_profile.renderingIntent = INTENT_SATURATION;
    } else if (strcasecmp(doc->header.cupsRenderingIntent, "Absolute") == 0) {
      doc->colour_profile.renderingIntent = INTENT_ABSOLUTE_COLORIMETRIC;
    }
```

surajkulriya:
Yes sir, I too was thinking the same. In fact thing you mentioned seems better then first one cause we are accounting for user supplied options too in that case. Will do this.

surajkulriya:
Sorry, we are accounting for user supplied options in first case too using cupsRasterParseIPPOptions and code written below. But anyway I will do the thing you mentioned
thanks.

PR #415, opened 2021-08-24
print-render-supported in pdftoraster filter function via already def…

surajkulriya:
…ined function


surajkulriya:
@tillkamppeter  sir, is this fine ?

tillkamppeter:
Thank you very much.

PR #421, opened 2021-09-04
Add a universal filter function to replace chain of individual filters

pranshukharkwal:
Submission for GSoC 2021 project : Create a universal filter function.
[Code Documentation + project wrap up blog](https://gist.github.com/pranshukharkwal/9413499a6744049ef549159948392023)


pranshukharkwal:
@tillkamppeter @dheeraj135

surajkulriya:
@tillkamppeter  sir @pranshukharkwal  has written really great code. Well, just for the sake of curiosity can any of you tell me why we are not using the weighted graph approach as we do in `cupsfilter` function of cups repository. 
Thank you

tillkamppeter:
Looks all nice, but one thing is missing: You have to `free()` all memory which you have allocated with `malloc()` and similar functions, especially as your filter function cannot be only called as a separate executable by CUPS but also can be called as a library function, for example from within a Printer Application. There the allocated and not freed memory would accumulate with each job.

tillkamppeter:
@surajkulriya Here we did not the weighted graph approach as we have a fixed set of filters (the ones in libcupsfilters) and we are always (with few exceptions) following the PDF-centric approach of CUPS printing, turning non-PDF input into PDF, applying `pdftopdf()`, turning the output into the format the PPD file or printer driver filter asks for. The approach of directly selecting filter functions for each input and output format makes it easier to control the filter chain, than having to adjust the cost values to arrive at the desired result.

pranshukharkwal:
@tillkamppeter I have added a commit for freeing the dynamically allocated memory.

tillkamppeter:
Could you also add a file with the MIME conversion rules, like `mime/universal.convs` with the MIME conversion rules for this filter? Do not add the rules to `mime/cupsfilters.convs` so that the user/admin can decide by installing the `universal.convs` file whether he wants to use the `universal` filter or not.

pranshukharkwal:
@tillkamppeter I have added the universal.convs and the universal.types files

tillkamppeter:
I have merged your work now, thank you very much.

dheeraj135:
Great work Pranshu! 😃

PR #425, opened 2021-09-17
texttotext converted to filter function

pranjanpr:


pranjanpr:
Added requested changes.

Issue #419, opened 2021-08-31
linking error


andyrtr:
I'm trying to build from master (84752a7e5f94cd848ad37b28b7e36904dc2daac6) "make install" fails here using Arch Linux:

```
configure:
==============================================================================
Environment settings:
	CFLAGS:                                    -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -Wall -std=gnu11 -D_GNU_SOURCE
	CXXFLAGS:                                  -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -Wall  -D_GNU_SOURCE
	LDFLAGS:                                   -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now
Build configuration:
	cups-config:     system
	font directory:  /etc/fonts/conf.d
	foomatic:        yes
	init directory:
	cups dom socket: /var/run/cups/cups.sock
	poppler:         yes
	ghostscript:     yes
	gs-path:         system
	mutool:          yes
	mutool-path:     system
	ippfind-path:    system
	imagefilters:    yes
	jpeg:            yes
	pdftocairo-path: system
	pdftops:         hybrid
	pdftops-path:    system
	png:             yes
	php:             no
	php-config:      system
	pstops:          no
	rastertopwg:     no
	shell:           /bin/bash
	test-font:       /usr/share/fonts/TTF/DejaVuSans.ttf
	tiff:            yes
	avahi:           yes
	dbus:            yes
	browsing:        DNSSD,CUPS
	werror:          no
	braille:	 yes
	braille tables:  /usr/share/liblouis/tables
	driverless:      yes
	apple-raster:    rastertopdf
	pclm:            yes
	local queue naming for remote CUPS queues: DNS-SD
	keep generated queues during shutdown:     no
	all ipp printer auto-setup: yes
	only driverless auto-setup: no
	only local auto-setup: no
==============================================================================
...
==> Starting package()...
 /usr/bin/mkdir -p '/build/cups-filters/pkg/cups-filters//usr/lib'
 /bin/sh ./libtool   --mode=install /usr/bin/install -c   libppd.la libcupsfilters.la libfontembed.la '/build/cups-filters/pkg/cups-filters//usr/lib'
libtool: install: /usr/bin/install -c .libs/libppd.so.1.0.0 /build/cups-filters/pkg/cups-filters//usr/lib/libppd.so.1.0.0
[...]

PR #427, opened 2021-09-21
Fix issues with build described in 419

nielsenb-jf:
Add missing line continuation character in libcupsfilters-la_LIBADD
stanza. This fixes `make[2]: lm: No such file or directory` errors while
building.

Move fontembed library before fontembed library, this fixes
`/usr/bin/ld: cannot find -lfontembed` errors during install. This may
just be a workaround for dependency resolving issues elsewhere.

[0] - https://github.com/OpenPrinting/cups-filters/issues/419

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -218,6 +218,70 @@ EXTRA_DIST += \
 	ppd/test2.ppd \
 	ppd/README.md
 
+# =================
+# Fontembed library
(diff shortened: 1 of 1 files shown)


tillkamppeter:
Thanks for the PR. I do not see any problems with it and as it solves your problems I have merged it.

PR #431, opened 2021-11-26
For some unknown reason, every printing file need sleep a little time…

fengyuewuhen:
For some unknown reason, every printing file need sleep a little time on first transmit
to make sure the serial printer receive data is right.


tillkamppeter:
@fengyuewuhen thanks a lot for your investigations and the resulting change, I have merged it now.

PR #433, opened 2021-12-02
driverless: Always use encryption for ipps

andrzej-kaczmarek:
RFC7472 requires that 'ipps' must be used over HTTPS, but driverless
does not enforce encryption and will use insecure connection by
default. This makes it unusable with a printer that expects secure
connection only on 'ipps' port.


andrzej-kaczmarek:
FYI, I noticed this problem when trying to add Canon MF744C as driverless. It announces both ipp and ipps, and I see ipps only discovered "by default". The printer then just ignores Get-Printer-Attributes request sent without encryption on ipps port.

tillkamppeter:
Merged, thank you very much for this patch.

PR #442, opened 2022-01-08
removing dejavu -dependency

### LINKED ISSUE #411: 1.28.10 still depends on DejaVuSans.ttf
According to `NEWS`
```
CHANGES IN V1.28.10
...
        - libfontembed: Removed unneeded fontembed/main.c and ttfread
          executable. Eliminates the dependency on DejaVuSans.ttf
          (Issue #386).
```
but `configure.ac` still contains
```
AS_IF([test "x$cross_compiling" != "xyes" && ! test -f "$with_test_font_path"],
        [AC_MSG_ERROR(DejaVuSans.ttf font file is missing. Please install a package providing it.)]
)
```
so cups-filter 1.28.10 can't be built without DejaVuSans.ttf.

khatri-deepak:
Removing dejavu- dependency from cups filter
fixes #411
changing error message to a warning so that cups filter can be installed without deja vu

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -790,7 +790,7 @@ AC_ARG_WITH([test-font-path],
 )
 
 AS_IF([test "x$cross_compiling" != "xyes" && ! test -f "$with_test_font_path"],
-	[AC_MSG_ERROR(DejaVuSans.ttf font file is missing. Please install a package providing it.)]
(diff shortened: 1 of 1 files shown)


tillkamppeter:
As your change in both master (2.x) and 1.x is the same I have applied the standard method of cherry-picking the change from master into 1.x, to express that the change is actually the same. In such cases it is enough to post a Pull Request for master.

But thank you anyway for the pull requests.

Issue #365, opened 2021-04-27
pdftopdf: "-o number-up" + "-o page-ranges"


xwitty:
Hi,

I am using cups on a
Raspberry Pi 2 Model B Rev 1.1
debian package cups-filters 1.21.6-5+rpt1

It seems, that the pdftopdf filter works correctly with either page-ranges or number-up but not with both!
The resulting PDF has an empty page tree.

what else can I provide as input?

thanks
michael

debiantriage:
cups-filters 1.21.6-5  is what comes with Debian buster.

I have just tested with `/usr/sbin/cupsfilter -p canonts6000.ppd -m application/vnd.cups-pdf test.pdf  -o number-up=2 -o page-ranges=1-2 > out.pdf`

out.pdf shows what I expect: two sheets with two pages from test.pdf on each sheet.

tillkamppeter:
@debiantriage Does this mean that the bug is fixed in the current version of cups-filters?

debiantriage:
@tillkamppeter  I used 1.28.7-1 from unstable. The outcome was no different from before. I will install 1.28.8-1: from experimental, if necessary.

xwitty:
thanks for your comments

I found the cause of the error!

2 examples on a source PDF with 16 pages printing with options:
-o number-up=2-o page-ranges=7-8
=> will print PDF pages 13-16 on 2 pages

-o number-up=2
-o page-ranges=15-16
=> will print nothing (empty page tree), as PDF has not enough pages

=> when using "number-up=" it is crucial to understand the sequence of processing (as the sequence of the options on the command line does not matter)
option A)
1. "number-up="
2. "page-ranges=" 
option B)
1. "page-ranges=" 
2. "number-up="

with option A, the page-range is applied to the already "number-uped (shrinked)" PDF
with option B, the page-range is applied to the original PDF, and "number-uped (shrinked)" is done in the 2nd step ...

obviously (at least my version) operates based on option A, but I used "page-ranges" like in option B (which might lead to page-ranges exceeding the number of pages in the "number-uped (shrinked)" PDF).

... if this works as intended. this issue should be closed
... if option B seems closer to what a user expects, then the processing sequence should be changed
for me option B makes more sense!

tillkamppeter:
This is now also added to `pstops()`, so that `input-page-ranges` is also available if we print PostScript from Adobe Reader to a PostScript printer, where we cannot convert the incoming PostScript to PDF.

[...]

PR #444, opened 2022-01-25
Added option input-page-ranges in pdftopdf filter

vermamohit13:
Added orig-page-range option in pdftopdf filter which will provide more flexibility to the user for printing pages in a certain range. Unlike the existing page-range option, which applies the page-range parameter to the document after it has been processed, the orig-page-ranges option applies page-range before the document has been processed.
This option will also help in the solving the need of user as mentioned in issue #365.

### DIFF
--- a/cupsfilters/pdftopdf/pdftopdf_processor.cc
+++ b/cupsfilters/pdftopdf/pdftopdf_processor.cc
@@ -32,14 +32,7 @@ bool ProcessingParameters::withPage(int outno) const // {{{
 // }}}
 bool ProcessingParameters::havePage(int pageno) const
 {
-  if (pageno%2 == 0) { // 1-based
-    if (!evenPages) {
-      return false;
-    }
-  } else if (!oddPages) {
-    return false;
-  }
-  return origPageRange.contains(pageno);
+  return inputPageRange.contains(pageno);
 }
 
 void ProcessingParameters::dump(pdftopdf_doc_t *doc) const // {{{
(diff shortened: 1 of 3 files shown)


cupsfilters/pdftopdf/pdftopdf_processor.cc
```diff
@@ -30,6 +30,17 @@ bool ProcessingParameters::withPage(int outno) const // {{{
...
+{
+  if (pageno%2 == 0) { // 1-based
+    if (!evenPages) {
+      return false;
+    }
+  } else if (!oddPages) {
+    return false;
+  }
+  return origPageRange.contains(pageno);
+}
```
tillkamppeter, line 44: Note that page-set=even and page-set=odd applies to the output pages as the original page-ranges attribute. The page-set attribute is for manual duplex (print odd pages, turn over the stack, put back into input tray, and print even pages). orig-page-ranges should simply select the pages from the input stream and the resulting list of pages passed on for all the other functionality of pdftopdf() to be applied.
If you have a 20-page input document and apply "orig-page-ranges=1-5,7-8,13-15 number-up=2 page-set=odd page-ranges=1-4" you get 2 sheets, the first contains the original pages 1 and 2, the second contains the original pages 5 and 7. Note that page-set and page-ranges are applied simultaneously. The page must be selected by both page-set and page-ranges to get finally selected. In our example we select 10 pages from the original 20 pages, then we put two on each sheet, ending up with 5 sheets, containing 1+2, 3+4, 5+7, 8+13, 14+15, after that we want to have the sheets 1, 2, 3,, and 4 but only the odd ones, so it remains sheet 1 and sheet 3, containing 1+2 and 5+7.

tillkamppeter:
Another point: Let us rename from orig-page-ranges to input-page-ranges, as we treat the input pages which go into pdftopdf().

cupsfilters/pdftopdf/pdftopdf_processor.cc
vermamohit13: Thanks for bringing this to my attention.I made the necessary adjustments, and the option now functions as you indicated.

vermamohit13:
> Another point: Let us rename from orig-page-ranges to input-page-ranges, as we treat the input pages which go into pdftopdf().

I have changed the name of the option to input-page-ranges.

tillkamppeter:
@vermamohit13 thanks a lot, it works perfectly for me. An important missing feature is added now.

PR #445, opened 2022-01-29
Added input-page-ranges to pstops filter

vermamohit13:
The input-page-ranges option is added to pstops filter.This option is implemented for the PostScript coming from the Adobe Reader which overrides the normal PDF workflow & pdftops.Hence, resulting in input-page-ranges(Implemented in #444) not being applied to the job. With this implementation the option will be applied to the postscript generated by Adobe Reader.

### DIFF
--- a/cupsfilters/pstops.c
+++ b/cupsfilters/pstops.c
@@ -106,6 +106,7 @@ typedef struct				/**** Document information ****/
 		page_border;		/* doc->page_border around pages */
   const char	*page_label,		/* page-label option, if any */
 		*page_ranges,		/* page-ranges option, if any */
+    *inputPageRange, /* input-page-ranges option, if any */
 		*page_set;		/* page-set option, if any */
   /* Basic settings from PPD defaults/options, global vars of original pstops */
   int           Orientation,            /* 0 = portrait, 1 = landscape, etc. */
@@ -150,6 +151,7 @@ typedef struct				/**** Document information ****/
 
 static pstops_page_t	*add_page(pstops_doc_t *doc, const char *label);
 static int		check_range(pstops_doc_t *doc, int page);
+static int		input_range_check(pstops_doc_t *doc, int page);
 static void		copy_bytes(pstops_doc_t *doc,
 				   off_t offset, size_t length);
 static ssize_t		copy_comments(pstops_doc_t *doc,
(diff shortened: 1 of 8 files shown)


cupsfilters/pstops.c
```diff
@@ -477,29 +586,31 @@ add_page(pstops_doc_t *doc,		/* I - Document information */
 
 static int				/* O - 1 if selected, 0 otherwise */
 check_range(pstops_doc_t *doc,		/* I - Document information */
```
tillkamppeter, line 588: Does not need "doc" in the call scheme any more, only page, Ranges, and pageset.

vermamohit13: Thanks for pointing this out.I have done the necessary change.

tillkamppeter:
Thanks, this was already a good first preparing step and a working model, now an important point for finalizing it:

The pstops() filter function needs to be completely streamable, as it is used in the streaming printing mode of Printer Applications (to print hige jobs on small devices without needing to hold the whole job, or to print infinite jobs), so we cannot copy the whole job into a temporary file. So please use the approach which I suggested, of creating a fork to filter out the unwanted pages according to input-page-ranges, like it is done in pdftops().

One fork does the input-page-ranges, a second does the original task of pstops() and the parent is feeding the input into the first fork. The first fork is only created if input-page-ranges is supplied.

cupsfilters/pstops.c
```diff
@@ -419,6 +494,20 @@ pstops(int inputfd,         /* I - File descriptor input stream */
...
+    { 
+      retry_wait:
+        if (waitpid (pstops_pid, &childStatus, 0) == -1) {
+          if (errno == EINTR)
+            goto retry_wait;
+          if (log) log(ld, FILTER_LOGLEVEL_ERROR,
+          "pstops: Error while waiting for input_page_ranges to finish - %s.",
+          strerror(errno));
+        }
+        if(log) log(ld, FILTER_LOGLEVEL_DEBUG, "pstops: input-page-ranges completed, status: %d", childStatus);
```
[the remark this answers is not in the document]
tillkamppeter: I have simply copied that from another filter function.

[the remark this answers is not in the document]
vermamohit13: Yeah Sure.Thanks.

cupsfilters/pstops.c
```diff
@@ -504,17 +505,34 @@ pstops(int inputfd,         /* I - File descriptor input stream */
...
+  }
 
   cupsFileClose(inputfp);
   close(inputfd);
 
   fclose(outputfp);
   close(outputfd);
 
-  return (0);
+  return (status ? 1:0);
```
tillkamppeter, line 535: As it is now with `status`, you can simply put `return (status);` here.

vermamohit13: OK, I have changed it.

Issue #362, opened 2021-04-25
-o print-scaling=none behaviour


debiantriage:
I attach the following two files:

![yellow](https://user-images.githubusercontent.com/19552178/115992885-de34a880-a5c7-11eb-9921-e2808200b627.jpg)

![garden](https://user-images.githubusercontent.com/19552178/115992962-408da900-a5c8-11eb-98e9-464a974ad82a.jpg)

The first is a scan taken with Xsane.  The second is from my camera.

`identify -v <JPEG_file>` shows 

> Print size: 6.53333x8.95333

> Print size: 9.16667x6.08333 

I would see both images as being smaller than A4. Executing

`/usr/sbin/cupsfilter -p rx420.ppd -m application/pdf -o PageSize=A4 -o print-scaling=none <JPEG_file> > out.pdf`

it seems to me that the fit method has been used to produce out.pdf. I would have expected a centred  image with four white borders. That is what I get with the image in Issue #65.

cups-filters is 1.28.7-1 on Debian unstable.

Regards,

Brian.

debiantriage:
The second image has resolution 192x192 according to `identify`. The output of the previous `cupsfilter` command shows 128x128 PPI. So it would seem that `imagetopdf` cannot get the resolution from the file and the image size is seen as being larger than A4. `-o print-scaling=auto` uses the expected fill method.

Being unable to obtain the resolution from the file is an extra complication. I imagine there may be good reasons why it cannot easily be accomplished.

Logs for processing both files have 

> [DEBUG]: Ignoring crop-to-fit option!

and in Issue #92 we see 

> For images, I am checking if more than 50% of height or width is to be cropped in crop-to-fit, then we ignore it.

However, in Issue #108 it is said that for `-o print-scaling=none`

> do not scale the document to fit the requested media size. If the document is larger than the requested media, center and clip the resulting output. If the document is smaller than the requested media, center the resulting output.

No limitation is placed on the degree of cropping that can take place.

The first image has its correct resolution identified, so it would appear `imagetopdf` should not be scaling it to fit A4.

Rishabh-792:
@debiantriage Can you please send the PPD file for this in a '.txt' format.

debiantriage:
Here you are: 
[rx420.ppd.txt](https://github.com/OpenPrinting/cups-filters/files/7877160/rx420.ppd.txt)

tillkamppeter:
[...]

PR #446, opened 2022-02-04
Changed the default ppi from 128 to 200.

Rishabh-792:
Fixes part of #362 .

Changed the default DPI resolution to 200 as suggested in the issue.


tillkamppeter:
I have merged it. Thanks.

PR #466, opened 2022-05-11
Adding exif reader

thakan25:
As a follow-up to #362, this patch makes use of libexif library to read EXIF metadata(xppi and yppi more specifically) from jpeg, png, and tiff images. If metadata is not present or we face an error, we fall back to either header or default value of 200dpi.

Implementation of EXIF reader function is called inside reader function of each of image type mentioned above. So every imageto.. type of filter will make use of the new implementation.


tillkamppeter:
Thanks for the PR

Seems all OK, but there are two things which needs to get fixed:
1. Do not add the file `mupdftopwg` to the repository. It is machine-generated by libtool
2. Merge your 3 commits into one, as the mess-up of indentation hides your actual changes

tillkamppeter:
Next attempt:
1. There are still a lot of pure white-space changes (former messing up and fixing of indentation did not match exactly). Please let your commit not contain pure white-space changes. 
2. `mupdftopwg` still gets added. Please do not add it.
3. In `Makefile.am` link the EXIF library only with libcupsfilters. The applications, filters, ... which link libcupsfilters do not to link the EXIF library by themselves.
4. Add a command line option for `./configure` so that the user can decide whether he wants to build with EXIF readout support or not.

Could you do these changes. If it gets too complicated you could also create a new PR which replaces this one (I will merge only the new PR then).

.gitignore
```diff
@@ -71,6 +71,7 @@ gstoraster
 imagetopdf
 imagetoraster
 mupdftoraster
+mupdftopwg
```
tillkamppeter, line 74: I have renamed `mupdftoraster` into `mupdftopwg` some weeks ago, as MuPDF (`mutool`) only outputs PWG Raster, no CUPS Raster and no Apple Raster. I simply forgot that I also had to update `.gitignore`. So could you remove `mupdftoraster` here to make it complete?

tillkamppeter:
Could you actually remove (`git rm`) the mupdftopwg file as the `.gitignore` addition does not do this automatically.

 Could you also do (3) and (4) of my comment above?

.gitignore
thakan25: Sure

tillkamppeter: Thanks.

Makefile.am
```diff
@@ -972,9 +974,11 @@ imagetopdf_SOURCES = \
...
 imagetopdf_CFLAGS = \
 	$(CUPS_CFLAGS) \
+	$(EXIF_CFLAGS) \
 	-I$(srcdir)/cupsfilters/
 imagetopdf_LDADD = \
 	$(CUPS_LIBS) \
+	$(EXIF_LIBS) \
 	libcupsfilters.la
 
 imagetops_SOURCES = \
```
tillkamppeter, line 984: You do not need this here for the individual filter executables, it is only needed in libcupsfilters itself, as you correctly did.  So remove these additions here.

Makefile.am
```diff
@@ -1199,6 +1204,7 @@ cups_browsed_LDADD = \
 	$(CUPS_LIBS) \
 	$(AVAHI_LIBS) \
 	$(AVAHI_GLIB_LIBS) \
+	$(EXIF_LIBS) \
 	$(GLIB_LIBS) \
 	$(GIO_LIBS) \
 	$(GIO_UNIX_LIBS)
```
tillkamppeter, line 1210: Also here the additions for libexif are not needed ...

cupsfilters/image.c
```diff
@@ -871,3 +874,121 @@ get_tile(cf_image_t *img,		/* I - Image */
...
+
+/*
+  implementation for EXIF read function
+  */
+
+/*
+  helper function to extract bytes from image files
+  */
+
+char *findBytes(FILE *fp, int *size)
```
tillkamppeter, line 901: Use `static char find_bytes(...)` here, as this function is local to this file. Also add this function to the overview comment at the top of the file.

cupsfilters/image.c
```diff
@@ -871,3 +874,121 @@ get_tile(cf_image_t *img,		/* I - Image */
 
   return (ic->pixels + bpp * (y * CF_TILE_SIZE + x));
 }
+
+/*
+  helper function required by EXIF read function
+  */
+
+void trim_spaces(char *buf)
```
tillkamppeter, line 882: Use `static void trim_spaces(...)` here, as this function is local to only this file.

cupsfilters/image.c
tillkamppeter: Especially also note our naming policy, file-local functions are named in all lower case with `_` as separator.

tillkamppeter:
Also add a command line option to `./configure` (via `configure.ac`) that the user can decide whether to use libexif or not. By default it should get used if it (and its headers) are present. Also do not forget to add a line to the overview in the end of `configure.ac` that one can see whether cups-filters will be built using libexif or not.

thakan25:
> 4\. Add a command line option for `./configure` so that the user can decide whether he wants to build with EXIF readout support or not.

Hi, Till. I am ready with this change, but still not aware why we have additional flags like AVAHI_GLIB_LIBS and AVAHI_GLIB_CFLAGS. I have not yet considered them for exif library. If you can provide some insight on why we have these flags it will help me include them for exif as well(if necessary).

Makefile.am
thakan25: Sorry Till, the changes you reviewed are old ones, I am working on a separate branch to resolve these issues and push the same and raise a pr soon

cupsfilters/image.c
thakan25: I should not declare local functions in image-private header file. right?

tillkamppeter: Functions which are used only in a single file, in the file where they are defined, do not need to be mentioned in any `*.h` file. The `...-private.h` files are for functions which are not part of the API of the library but which are used in more than one source file of the library. So `trim_spaces()` does not need to be put into any `*.h` file.

thakan25: Yeah, got it

thakan25: I have made desired changes in commits below. Please have a look.

tillkamppeter:
Now all is correct.

I have merged your work now. Thank you very much.

PR #477, opened 2022-08-16
Moves the libppd declarations after cups-filters in Makefile.am

CarlosNihelton:
When compiling from source `sudo make install` fails to relink libppd because it depends on libcupsfilters, not yet installed.
Reordering the declarations in the Makefile.am and correcting the assignments of lib_LTLIBRARIES, check_PROGRAMS and TESTS fix the behavior.

To reproduce the issue try on a minimal Ubuntu environment, such as WSL or cloud images running on LXD or docker:

```
sudo apt install autotools-dev autoconf automake autopoint libavahi-client-dev libavahi-glib-dev libcurl4-gnutls-dev libglib2.0-dev libgnutls28-dev libjpeg-dev libpam0g-dev libpng-dev libusb-1.0-0-dev zlib1g-dev libcups2-dev libcupsimage2-dev libssl-dev libtool libqpdf-dev poppler-utils libfontconfig1-dev libpoppler-cpp-dev libtiff5-dev libdbus-1-dev libexif-dev librsvg2-bin sharutils liblcms2-dev ghostscript pkg-config g++ --no-install-recommends

git clone https://github.com/OpenPrinting/cups-filters
cd cups-filters
./autogen.sh
./configure --prefix=/usr --disable-ldap --disable-foomatic --disable-braille --disable-mutool
make
sudo make install
```

Only the install target should fail to relink the libppd against libcupsfilters due the later being not yet installed.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -161,181 +161,6 @@ pkgcharset_DATA = \
 
 EXTRA_DIST += $(pkgcharset_DATA)
 
-# ==========================
-# PPD legacy support library
-# ==========================
-pkgppdincludedir = $(includedir)/ppd
-pkgppdinclude_DATA = \
-	ppd/ppd.h \
-	ppd/ppdc.h \
-	ppd/ppd-filter.h
-
-pkgppddefsdir = $(datadir)/ppdc
-pkgppddefs_DATA = \
-	ppd/epson.h \
-	ppd/hp.h \
-	ppd/label.h \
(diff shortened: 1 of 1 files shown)


tillkamppeter:
I have merged it now, thanks a lot for finding this out and testing the fix. I really did not know that the order of the rules in a Makefile matters, was always thinking that `make` finds the right order by itself ...

PR #482, opened 2022-09-15
Add FrequentNetifUpdate option/directive

zdohnal:
In case clients use BrowsePoll to a server with many queues or there is many printers on the local network, updating network interfaces after every found record slow cups-browsed a lot (f.e. on Fedora this step takes about 2-3s second for every queue). This PR adds FrequentNetifUpdate configure option and directive, which can turn off this update when we examine found records - the network interface update based on NM DBUS notification is not affected with this directive.

Thank you in advance!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -962,6 +962,19 @@ if test x$enable_debug_guards = xyes; then
         CXXFLAGS="$CXXFLAGS -DDEBUG_GUARDS"
 fi
 
+# =========================================================
+# Turn on/off network interface updates for each found entry
+# ==========================================================
+AC_ARG_ENABLE([frequent_netif_update],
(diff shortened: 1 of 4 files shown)


tillkamppeter (approved):

tillkamppeter:
That is a good idea. Thank you very much.

tillkamppeter:
Better would even be not only having "On" and "Off" but also an automatic mode, which does frequent network interface updates if there are 0-5 queues and not if there are more queues. This could be the default in distros. Then for small-network users all works automatically and who has larger networks is rather enterprises which have an admin can easily cope with extra work like restarting cups-browsed.

zdohnal:
@tillkamppeter that's a great idea! I've designed something similar for timeouts in cups-browsed, but haven't got into it yet, it is on TODO list (since the original issue why I came up with timeout calculaction was solved by other way).

PR #504, opened 2023-02-02
configure.ac: C++ not needed

zdohnal:


tillkamppeter:
Thank you very much, @zdohnal.

Issue #480, opened 2022-09-01
cups-filters 1.28.16: build fails, libexif not found


Vouivre:
I updated `cups-filters` 1.28.15 to 1.28.16, but I get the following
when I build it on my linux system:

```
checking for library containing TIFFReadScanline... -ltiff
checking for libexif... no
configure: error: Package requirements (libexif) were not met:

Package 'libexif', required by 'virtual:world', not found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.
```

So it seems now `libexif` is required, unless I provide 

    --disable-exif

in the configure script. I assume it was introduced with 
https://github.com/OpenPrinting/cups-filters/issues/362

I had a look in the README and I didn't see that `libexif` is required to
build `cups-filters`. Shouldn't it be added to the README ?

tillkamppeter:
It should be mentioned in `INSTALL` ...

PR #511, opened 2023-02-21
Add libexif in README.md

srohatgi01:
As mentioned in #480, from cups-filters 1.28.16, `libexif` is now required and is now mentioned in the README.md as well.

In libcupsfilters's INSTALL, it's already mentioned that `libexif` is required to get the image resolution.

I am new to this so please let me know if there's anything.

Thanks in advance!

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ package. This package retro-fits the filter functions of
 libcupsfilters and libppd to CUPS 2.x.
 
 For compiling and using this package CUPS (2.2.2 or newer),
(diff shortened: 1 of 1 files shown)


tillkamppeter:
No need for explicit mention of libexif in `README.md`, as we already require libcupsfilters and libppd, and we do not mention libjpeg, libpng, QPDF, and anything else needed by libcupsfilters.

But thank you anyway for the PR.

srohatgi01:
What shall be done for the issue #480?

PR #555, opened 2023-10-25
braille: Fix finding tables from liblouis ≥3.21

sthibaul:
Since 3.21 (more precisely, commit e18fe140e2f1), liblouis uses language: and region: rather than locale:


tillkamppeter:
@sthibaul there is a conflict when merging. Could you check and update your changes? Thanks.

sthibaul:
It was actually the same as https://github.com/OpenPrinting/cups-filters/pull/590

PR #558, opened 2023-11-09
configure.ac: remove final traces of avahi

### LINKED ISSUE #556: Remove avahi tests in configure
If avahi is not installed, and `--disable-avahi` is not passed, `configure` errors out. But this shouldn't happen, since neither `AVAHI_CFLAGS` nor `AVAHI_LIBS` nor `HAVE_AVAHI` is used in the code...

1. Uninstall avahi-dev if it is installed
2. run configure without passing `--disable-avahi`
3. It stops with a message starting with: "configure: error: Package requirements (avahi-client) were not met ..."
 
**Expected behavior**
Since avahi is nowhere used, it shouldn't be needed to have configure run to completion.

**System Information:**
 - reproducible on both debian bookworm and linux from scratch. cups-filters version is 2.0.0

eli-schwartz:
This was originally searched for cups-browsed, and during the great split these lookups accidentally got left in and moved around in the file.

The original code began with this autoconf comment:
```
dnl Avahi for cups-browsed
```

After shifting down, it became:
```
dnl Avahi
```

After removing avahi from configure.ac, there are no remaining mentions of avahi anywhere other than CHANGES*.md

Fixes #556

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -268,32 +268,6 @@ AS_IF([test "x$with_ippfind_path" != "xsystem"], [
 AC_DEFINE_UNQUOTED([CUPS_IPPFIND], "$CUPS_IPPFIND", [ippfind binary to use.])
 
 
-dnl Avahi
-AVAHI_LIBS=""
-AVAHI_CFLAGS=""
-
-AC_ARG_ENABLE([avahi],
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Gentle ping. :)

Issue #541, opened 2023-09-13
lp -o page-set=even prints an extra blank page


olivluca:
I'm using my forked [manual_duplex_linux](https://github.com/olivluca/manual_duplex_linux) repository to print in manual duplex mode: it first print the odd pages then the even ones (in reverse order and rotated 180º to cater to the paper flow in my printer), so the option are `-o page-set=even -o collate=true -o outputorder=reverse -o orientation-requested=6`
When the document has an odd number of pages, everything is fine, but if it has an even number of pages, an extra blank page is printed which messes up the order.
If I run the pdftopdf filter with those same parameters, no extra blank page is generated, so I don't know where that extra page comes from.
For the time being, I use pdftk to filter and rotate the even pages if the number of pages is even and lp just print them in reverse order.
This is on debian 12 with cups-filters 1.28-17-3, but I have the same problem on another computer still running ubuntu 18.

tillkamppeter:
Could you attach the PPD file of your print queue (from `/etc/cups/ppd/`)? Please rename it to have the name extension `.txt` so that GitHub accepts it.

olivluca:
Sure, here it is
[Manual_Duplexer_Pantum_M6500W_series_4CF883.txt](https://github.com/OpenPrinting/cups-filters/files/12678105/Manual_Duplexer_Pantum_M6500W_series_4CF883.txt)

This is another one using a virtual pdf printer (so I don't waste paper while doing tests)
[Manual_Duplexer_PDF.txt](https://github.com/OpenPrinting/cups-filters/files/12678110/Manual_Duplexer_PDF.txt)

kerneis:
For the record, I face the exact same issue using only the regular GUI printing interface. I'm printing to a USB Canon MF4700 (auto-detected by CUPS), and if I select "pages: odd" and "output order: reverse" in the print menu for a 10-page document, I end up with 6 printed sheets, the first one being blank. I have no issue when I select "pages: even" so my workaround for now is to start printing the odd pages (reversed), remove the extra blank page, and then feed back the result to the printer for the even pages — but it's annoying and can waste a lot of paper if I forget doing it.

I'd be happy to provide more information if you let me know what you need.

It looks like at least one other person faced the issue here: https://access.redhat.com/solutions/76333 but I don't have subscription so cannot read the solution they ended up with.

ValdikSS:
I can confirm this, and this is reproducible with cups-pdf.
[...]

PR #614, opened 2025-01-03
pdftopdf: do not generate extra blank page for page-set=even (#541)

ValdikSS:
Since commit 686a28d pdftopdf started to insert extra blank page when printing even-paged documents in page-set=even mode. This was the attempt to fix two issues at once:

* [Print 1-page documents in even mode][1]
* Treat non-duplex even reverse printing as second step of manual duplex

Unfortunately due to inverted logic this didn't fix any issues and introduced new one, outlined above.

This code fixes printing 1-page documents in even mode, and prints exactly what the user requested in even mode for larger documents, without inserting extra blank pages.

[1]: https://bugs.launchpad.net/ubuntu/+source/cups-filters/+bug/1340435


tillkamppeter:
Found the actual problem which was wrong counting of the pages. Now blankpages get added correctly. Clsing this PR. But thanks anyway.

PR #678, opened 2026-03-09
filter/pdftopdf: Add print-as-image option (issue #601)

### LINKED ISSUE #601: Offer a "Print as image" option for all the printers
**Is your feature request related to a problem? Please describe.**
A problem we hit from time to time is that, for some PDFs, our printer prints wrong characters or, even, nothing at all. This problem is workarounded choosing the "Print as image" option offered by the Google Chrome PDF viewer (the same exists for Adobe Reader on other platforms). Sadly, it is not widely implemented on other PDF readers

**Describe the solution you'd like**
It would be nice if, under the general printer options (those allowing to choose the paper size, duplex, error policy...) an option to "Print as image" would be available.

**Describe alternatives you've considered**
For now I rely on Google Chrome PDF reader as a workaround.

Thanks a lot

KarunishaR:
Fixes #601

Adds a new `print-as-image` job option to the pdftopdf filter. When
enabled, each PDF page is rasterized via Ghostscript's pdfimage24
device before printing, fixing printers that misrender vector PDF
content or fonts.

Usage:
  lpr -o print-as-image=true file.pdf
  lpr -o print-as-image=true -o print-as-image-dpi=600 file.pdf

Tested with Ghostscript 10.02.1:
- Rasterized output: 29K vs 2.3K original (confirms bitmap conversion)
- Fallback to original PDF works if Ghostscript fails
- Normal path without option works unchanged

### DIFF
--- a/filter/pdftopdf.c
+++ b/filter/pdftopdf.c
@@ -1,69 +1,122 @@
 //
 // Legacy CUPS filter wrapper for ppdFilterPDFToPDF() for cups-filters.
 //
-// Copyright © 2020-2022 by OpenPrinting.
+// Adds support for the "print-as-image" job option (issue #601):
+// rasterizes each PDF page via Ghostscript before processing, as a
+// workaround for printers that misrender PDF vector/font content.
+//
+// Copyright © 2020-2024 by OpenPrinting.
 //
 // Licensed under Apache License v2.0.  See the file "LICENSE" for more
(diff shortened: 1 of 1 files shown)


KarunishaR:
Hi, just following up on this PR. I’d appreciate any feedback or review when convenient. I’m happy to make any changes requested.

Issue #697, opened 2026-06-26
Potential NULL pointer dereference cases


dhejyuyr:
Hello! I found some issues via a SAST tool. They are not critical, but I hope this information could be useful.

There are two potential NULL pointer dereference cases:

1. filter/rastertopclx.c

`ppdLutLoad()` and `cfLutNew()` can return NULL if the nested `calloc()` function fails. There is no check on line 443:

https://github.com/OpenPrinting/cups-filters/blob/75e4bd9fa9da584948c8c00117a5d5542beb5b7b/filter/rastertopclx.c#L433-L443

Add a NULL check of `cfLutNew()` result.

2. filter/foomatic-rip/spooler.c

The `HOME` environment variable is usually defined in user environments, but this isn't guaranteed. 

https://github.com/OpenPrinting/cups-filters/blob/75e4bd9fa9da584948c8c00117a5d5542beb5b7b/filter/foomatic-rip/spooler.c#L229

Add a NULL check of `getenv("HOME")` result.

AayushKumar26:
Hello ! 
I noticed that PR #701 addresses the first case here (the` cfLutNew()` NULL check in `rastertopclx.c`), but the second case reported in this issue - the unchecked `getenv("HOME")` in` filter/foomatic-rip/spooler.c `(line 229) - isn't covered by it.
I'd like to work on that second case and will open a separate PR for it shortly.
One small note: since #701 is marked "Fixes #697," merging it would auto-close this issue while the `spooler.c` case is still unresolved. Might be worth keeping that in mind so the second case isn't lost.
Thanks!

PR #702, opened 2026-07-14
foomatic-rip: check getenv("HOME") result before use

AayushKumar26:
In init_direct() (filter/foomatic-rip/spooler.c), getenv("HOME") is passed straight into strlcpy() with no NULL check. When HOME is unset, getenv returns NULL and the filter crashes. This guards the result, falling back to the current directory and logging via _log(), matching the existing getenv() guard pattern already used at line 51 of the same file.
This addresses the second case from #697 (the foomatic-rip/spooler.c NULL deref), which is not covered by #701 (that PR handles the rastertopclx.c case). Using "Addresses" rather than "Fixes" so #697 isn't auto-closed while the other case is still open.

### DIFF
--- a/filter/foomatic-rip/spooler.c
+++ b/filter/foomatic-rip/spooler.c
@@ -225,8 +225,16 @@ init_direct(list_t *arglist,
   char tmp [1024];
   listitem_t *i;
   char user_default_path [PATH_MAX];
-
-  strlcpy(user_default_path, getenv("HOME"), 256);
+  
+  char *home = getenv("HOME");
+  if (home)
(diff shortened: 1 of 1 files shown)


tillkamppeter:
@AayushKumar26 could you remove the commented out original line? It makes the CodeQL test fail.

AayushKumar26:
> @AayushKumar26 could you remove the commented out original line? It makes the CodeQL test fail.

Hi @tillkamppeter , thank you for taking a look and for the quick review! 
You're right - I'll remove the commented-out line and push an updated commit shortly.

